PR #13, opened 2021-01-11
Add local config option

kylape:


bsquizz:
I've tested this, with the example conf, but none of those ClowdApps are using `dependencies` -- we just need to test that the logic for downloading dependencies is working

bsquizz (approved):

PR #14, opened 2021-01-11
Set up environment for Analytics

appuk:
Add config for Automation Analytics so that the iqe tests can target the right deployment


appuk:
@bsquizz let me know if this looks right/ if there is some way to test this

bsquizz (approved):

PR #22, opened 2021-01-28
skip reserving namespace if it is already owned_by_me

jameswnl:


bsquizz:
This should be addressed as part of https://github.com/RedHatInsights/bonfire/commit/27a1662b4971dd67e0eab7274362a18cdfcbca99

PR #23, opened 2021-02-18
Move local config to its own subcommand

kylape:


bonfire/bonfire.py
```diff
@@ -120,14 +126,30 @@ def _warn_if_not_available(namespace):
...
         "--app",
         "-a",
         "apps",
         required=True,
         help="comma,separated,list of application names",
     ),
+    click.option(
+        "--set-image-tag",
+        "-i",
+        help="Override image tag for an image using format '<image name>=<tag>'",
```
bsquizz, line 140: The `help` text isn't accurate for local mode :P

bonfire/bonfire.py
```diff
@@ -61,6 +61,12 @@ def config():
     pass
 
 
+@main.group()
+def local():
+    """perform operations related to local development"""
```
bsquizz, line 66: Perhaps more accurate is: "perform operations using a local config file"?

bsquizz (approved):

PR #25, opened 2021-02-18
Bonfire v1.0

bsquizz:
This PR brings a major rework of the CLI UX, adds some new enhancements for dealing with local configuration file, and refactors the template processing logic so that the same "processor" is used regardless of the source of the app configs (local vs app-sre)

You can test this out with: `pip install crc-bonfire==v1.0.0`

Changes for this PR:

*Changes to CLI*
* `bonfire config get` now becomes `bonfire process`
* `bonfire config deploy` and `bonfire local deploy` now becomes one command: `bonfire deploy`
* By default the "config source" used is `local`, but you can use `--source` to switch to appsre mode (see "Configuration files" section below)
* Added `bonfire process-env` and `bonfire deploy-env` which can be used to take a ClowdEnvironment template and apply it into a cluster. By default, the ephemeral template provided in the bonfire package is used.

*Changes to template processing*
* `--apps` option is now an argument: specify apps with spaces as `app1 app2 app3` or commas as `app1,app2,app3`
* Ability to switch config source between `local` or `appsre` using `--source`
* By default, we will set replicas to all ClowdApps to '1' unless you use `--no-single-replicas`
* By default, we will get dependencies for ClowdApps unless you use `--no-get-dependencies`
* By default, we will remove resource limits/requests from ClowdApps unless you use `--no-remove-resources`
* Added `--set-parameter` which uses format `<app name>/<component name>/<param name>=<value>`
* `--set-template-ref` uses format `<app name>/<component name>=<ref>`
* `--src-env` was renamed to `--target-env`
* The `process` options work the same way whether you are using `--source local` or `--source appsre` except for two less commonly used options:
`--target-env` (only applies for appsre configs)

*Changes to deploy behavior*
`bonfire deploy` now tries to infer information about what kind of cluster you're deploying to:
* we look for namespaces that are "reservable" (by getting all namespaces and checking for the labels that the namespace reconciler uses)... if we do find reservable namespaces, we will try to reserve one. If we cannot find any, you need to provide `--namespace`
* we look for a ClowdEnvironment on the cluster that is mapped to the target namespace. If one is not found, you need to provide the ClowdEnvironment name you want to map your apps to using `--clowd-env`, or when working on local minikube you can use `bonfire deploy-env`

*Configuration files*
* We now look for your env variable settings in `$XDG_CONFIG_HOME/bonfire/env`
* By default, we look for your config.yaml in `./config.yaml`, then `$XDG_CONFIG_HOME/bonfire/config.yaml`. You can still use `--local-config-path` if you want to use a totally different file.
* If you run `bonfire` and no local config.yaml exists yet, we copy a default config.yaml from this package into `$XDG_CONFIG_HOME/bonfire/config.yaml` for you. This will be used as our "recommended local config".
* The default configuration downloads app configurations from an internal location here: https://gitlab.cee.redhat.com/insights-platform/cicd-common/-/blob/master/bonfire_configs/ephemeral_apps.yaml -- we still have the `ephemeral deploy target` set up for apps in app-interface but we might phase that out in favor of using this file in future.
* You can still define `apps` in your config.yaml if you wish you add or override an app. An app defined in the `apps` section will override any remotely fetched config for that app (in both source modes)
* If you want to reset your config back to the "bonfire-supplied" one, use `bonfire config write-default`
* If you want to create a new config.yaml file in another location starting with the "bonfire-supplied" one as your template, use `bonfire config write-default <PATH>`. You could later run with `bonfire process -c <PATH> <app names>`
* The local app config format now is a list of apps. Each app can have 1 or more components. This aligns with the way things are set up in app-interface... so for example now `bonfire process ingress gateway insights-ephemeral --source local` can behave the same way as `bonfire process ingress gateway insights-ephemeral --source appsre`

*Examples*

To deploy onto local minikube using your local apps config.yaml:
```
kubectl create namespace test
bonfire deploy-env -n test
bonfire deploy app1 app2 app3 -n test
```

To deploy onto local minikube using AppSRE ephemeral configs:
```
kubectl create namespace test
bonfire deploy-env -n test
bonfire deploy app1 app2 app3 --source appsre -n test
```

To deploy onto ephemeral cluster using your local apps config:
```
bonfire deploy app1 app2 app3
```

To deploy onto ephemeral cluster using app SRE config:
```
bonfire deploy app1 app2 app3 --source appsre
```

To deploy onto ephemeral cluster using a namespace you've already reserved, specify the same namespace again:
```
bonfire deploy app1 app2 app3 -n YOUR-NAMESPACE
```

For running in pr_check.sh, the command previously looked like this:
```
bonfire config deploy \
    --ref-env insights-stage \
    --app $APP_NAME \
    --set-template-ref $COMPONENT_NAME=$GIT_COMMIT \
    --set-image-tag $IMAGE=$IMAGE_TAG \
    --get-dependencies)
```

It will now look like this:
```
bonfire deploy \
    $APP_NAME \
    --ref-env insights-stage \
    --set-template-ref $APP_NAME/$COMPONENT_NAME=$GIT_COMMIT \
    --set-image-tag $IMAGE=$IMAGE_TAG \
```


psav:
The description seems to add the functionality we discussed, I will try out the new branch today

bsquizz:
You can try it with: `pip install crc-bonfire==v1.0.0a0`

kylape:
* I think we should leave the example config file checked in to gitlab since it will refer to private repos if we want it to be a comprehensive list of apps
* Pretty sure it's a typo, but in your third bullet, instead of `bonfire config local`, did you mean `bonfire local get`?
* I think we should chose different names for the new hierarchy keys.  E.g. instead of "apps" and "components", how about "groups" and "apps"?  My concern is that calling the group of ClowdApps "apps" is confusing when the list of apps is called "components".

bsquizz:
> * I think we should leave the example config file checked in to gitlab since it will refer to private repos if we want it to be a comprehensive list of apps

Ok. I was hoping to keep it "pre-loaded" in the package -- think it'd be alright if we kept a pointer to the gitlab URL of the "internal config" in this repo? 

> * Pretty sure it's a typo, but in your third bullet, instead of `bonfire config local`, did you mean `bonfire local get`?
Yes, thanks, fixed.

> * I think we should chose different names for the new hierarchy keys.  E.g. instead of "apps" and "components", how about "groups" and "apps"?  My concern is that calling the group of ClowdApps "apps" is confusing when the list of apps is called "components".
Oof... Yeah, I was trying to use similar naming to the app-sre structure where an `app` has a list of `resourceTemplates`. For example in app-interface, the `ingress` app has 3 `resourceTemplates`: ingress, storage-broker, and puptoo. So I am reluctant to use `apps` in lieu of `components` because an "app" in app-sre means something different as well. What's a better word for the top level "thing"? :thinking:

bsquizz:
I was thinking about bonfire CLI some more. Based on the use cases I'm starting to see, wondering if something like this would be better:

`bonfire config get` -- does what it does now.
`bonfire local deploy` -- handles stuff to get config and deploy apps in "one command" on local kube
`bonfire remote deploy` -- handles all the stuff for you to get config and deploy an app in "one command" on the remote ephemeral cluster

In all 3, we would have a `--source` flag that lets you select from qontract or local (default could be local) -- this would change how bonfire looks up your app configs.

And then we would try to keep the behavior of the `--option` flags as similar as possible so that `bonfire config get` works almost the same way no matter which 'config source' you are using, and the command could be used interchangeably. So for ex `--set-image-tag`, `--set-template-ref`, etc. -- I'm thinking these should behave the same way no matter which "source of truth" you are using. We could add new flags that do "overrides" in different ways if we want, but try to keep their behavior the same for either config source.

kylape:
Why not put `get` under the `local` and `remote` subcommands:

`bonfire local get`
`bonfire local deploy`
`bonfire remote get`
`bonfire remote deploy`

This way the semantics are free to diverge (within reason), and the symmetry will hopefully make it more intuitive how the CLI is structured.

bsquizz:
To clarify what I'm saying is I'd like this kind of command to work no matter where I run it:

(let's say `-s` means `--default-source` -- i.e. the source we query first to get app configs out of.)
```
bonfire config get -s local -a someapp --set-image-tag <SOME SYNTAX> --set-template-ref <SOME SYNTAX> | kubectl apply -f -
```

Could behave the same as:
```
bonfire config get -s qontract -a someapp --set-image-tag <SOME SYNTAX> --set-template-ref <SOME SYNTAX> | kubectl apply -f -
```

The local config.yaml is basically just a replacement for app-interface. It has all the same information in it (your list of apps, your template refs, the params you process each template with) -- it is just formatted in a different way. So IMO options like `--set-template-ref` and `--set-image-tag` could behave the same way no matter what the config source is.

If we want to change the behavior of --set-image-tag to only update the IMAGE_TAG param for a single component then ok we can do that -- but let's just make it behave the same way for both config sources.

So if I have a namespace on minikube and I run that, it applies the same resources as if I had reserved a namespace on a remote cluster. 


What the `deploy` commands would do would be more of the "wiring together" -- i.e. `remote deploy` should reserve a namespace for you. `local deploy` might ensure you have a ClowdEnvironment provisioned, etc. -- and then once you have that namespace set up it would run essentially the same logic that `bonfire config get` runs and apply those resources into the destination namespace. But when it comes to actually processing the ClowdApp templates, I can see it being useful to have the 'config get' command working the exact same way no matter which cluster you're trying to apply a resource to. If someone runs into a problem on the ephemeral cluster, I could run `bonfire config get` with the same flags against my local minikube. `config get`'s job would be just to process your templates, and override some parameter/image/tag/etc. values in them if you want to -- but remain agnostic about WHERE you're actually applying those configs.

bsquizz:
Latest changes for v1.0 are ready for testing. See the PR description to see all of the changes, and try it out with:
```
pip install crc-bonfire==v1.0.0b0
```

bsquizz:
Latest changes in: `pip install crc-bonfire==v1.0.0b1`

bsquizz:
Latest changes in: `pip install crc-bonfire==v1.0.0b2`

bsquizz:
Released as: `pip install crc-bonfire==v1.0.0`

PR #31, opened 2021-03-22
Set NAMESPACE when deploy fails

psegedy:
NAMESPACE env is needed in teardown to collect logs and release reserved namespace even when deploy fails

why was namespace exported only when deploy succeeded? I hope that bonfire is returning exit code other than 0 when deploy fails (and jenkins stops), so removing this code won't break anything
```bash
if [ $? -eq 0 ]; then
    export NAMESPACE=$result
fi
```

### DIFF
--- a/cicd/deploy_ephemeral_env.sh
+++ b/cicd/deploy_ephemeral_env.sh
@@ -3,13 +3,11 @@ source ${CICD_ROOT}/_common_deploy_logic.sh
 # Deploy k8s resources for app and its dependencies (use insights-stage instead of insights-production for now)
 # -> use this PR as the template ref when downloading configurations for this component
 # -> use this PR's newly built image in the deployed configurations
-result=$(bonfire deploy \
+export NAMESPACE=$(bonfire namespace reserve)
+bonfire deploy \
     ${APP_NAME} \
     --source=appsre \
     --ref-env insights-stage \
     --set-template-ref ${APP_NAME}/${COMPONENT_NAME}=${GIT_COMMIT} \
-    --set-image-tag ${IMAGE}=${IMAGE_TAG})
-
-if [ $? -eq 0 ]; then
-    export NAMESPACE=$result
-fi
+    --set-image-tag ${IMAGE}=${IMAGE_TAG} \
+    --namespace ${NAMESPACE}


bsquizz:
`bonfire config deploy` is an "efficient 1-liner" that: reserves a namespace, gets the conigs, deploys the apps, waits for them to come up, and then finally when that is all done prints the namespace that was deployed to on STDOUT. The shell doesn't know what the namespace is until that command completes successfully. bonfire has a try/except which will release the namespace if the reserve/deploy fails:

https://github.com/RedHatInsights/bonfire/blob/master/bonfire/bonfire.py#L346-L351

If we want to collect logs in the shell, I think we need to re-write the shell script to do this:
 1. `export NAMESPACE=$(bonfire namespace reserve)`
 2. `bonfire config get ... | oc apply -f - -n $NAMESPACE`
 3. `bonfire namespace wait-for-resources $NAMESPACE`
 4. if 2 or 3 fails, collect logs, and call `bonfire namespace release $NAMESPACE`
 
 We might want to wait on #25 because I am reworking the CLI UX and the names of some commands are about to change

bsquizz:
(or, alternatively, we could start to add some code to pull logs/events out of the NS into bonfire...)

psegedy:
I did that because I noticed in a 1 failing jenkins job that it was killed right after `bonfire config deploy` failed during deploy phase, `teardown` ran, but it did nothing since the `NAMESPACE` was not set (it probably released the namespace, I don't remember whether there was anything in logs).

why do I need to do steps 2, 3 form your comment and I can't run
1. `export NAMESPACE=$(bonfire namespace reserve)
2. `bonfire config deploy --namespace $NAMESPACE
3. `bonfire namespace release $NAMESPACE` was already in `teardown`


Anyway, I agree it would be better to wait for #25

psegedy:
Based on the following code it won't release namespace when deploy fails and `requested_ns` is set, so the change in this PR should be enough to run teardown when deploy fails https://github.com/RedHatInsights/bonfire/blob/2bbd1d8639b1ff28b3787d7a9b619ea62b098239/bonfire/bonfire.py#L587-L591

bsquizz (approved):

PR #32, opened 2021-03-22
Fix missing set_image_tag argument

psegedy:
`bonfire local get --get-dependencies` fails when ClowdApp has dependencies because `processed_apps` is passed instead of `set_image_tag`

If I understood the code correctly `set_image_tag` is only used for `app` not for its `dependencies`. That's why I set it to `[]`


bsquizz:
There's a big re-write of all this logic in #25

PR #51, opened 2021-05-17
Remove duplicate short flag

patrickeasters:
This removes the misleading `-t` short flag for `--set-template-ref`. Because`timeout` already defines this flag, `-t` gets interpreted as `timeout` instead.

The `<app>/<component>=<ref>` syntax is also deprecated according to a further log message, so this updates the help text for this option as well.


bsquizz (approved):
Thanks!

PR #52, opened 2021-05-18
Allow setting different dockerfile name

psegedy:
Patchman team uses different Dockerfile name i.e. https://github.com/RedHatInsights/patchman-engine/blob/master/Dockerfile.rhel8

### DIFF
--- a/cicd/build.sh
+++ b/cicd/build.sh
@@ -2,6 +2,7 @@
 
 # Env vars caller defines:
 #IMAGE="quay.io/myorg/myapp" -- docker image URI to push to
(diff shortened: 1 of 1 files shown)


cicd/build.sh
```diff
@@ -15,6 +16,8 @@
 
 set -ex
 
+DOCKERFILE=${DOCKERFILE:="Dockerfile"}
```
bsquizz, line 19: Wasn't aware of the `:=` operator, nice.

bsquizz (approved):

PR #57, opened 2021-06-04
Check for marketplace host and port presence

catastrophe-brandon:
If the configuration is missing hostname or port values, this will cause subsequent code to fail with a syntax error in the f-string on line 68. To prevent this error, perform some basic checking to confirm the existence of hostname and port. If they are not configured, raise an error.

### DIFF
--- a/cicd/iqe_pod/create_iqe_pod.py
+++ b/cicd/iqe_pod/create_iqe_pod.py
@@ -61,6 +61,10 @@ def _build_test_conf(env_parser):
 
     if env_parser.app_present("marketplace"):
         mp_storage_cfg = env_parser.get_storage_config("marketplace")
+        if mp_storage_cfg.get('hostname', '') == '':
+            raise
+        if mp_storage_cfg.get('port', '') == '':
+            raise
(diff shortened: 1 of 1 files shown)


cicd/iqe_pod/create_iqe_pod.py
```diff
@@ -61,6 +61,10 @@ def _build_test_conf(env_parser):
 
     if env_parser.app_present("marketplace"):
         mp_storage_cfg = env_parser.get_storage_config("marketplace")
+        if mp_storage_cfg.get('hostname', '') == '':
```
catastrophe-brandon, line 64: The default value could be `None` just as easily as empty string.

bsquizz:
What was the traceback for the issue you ran into?

`mp_storage_cfg.hostname` and `mp_storage_cfg.port` should be `None` by default according to:
https://github.com/RedHatInsights/app-common-python/blob/master/app_common_python/types.py#L221-L225

and having `None` as an f-string appears to be valid:
```
>>> f"{None}"
'None'
```

PR #58, opened 2021-06-07
Add option to select components for app

psegedy:
`-C` or `--component` option to select which components should be processed. Set via `COMPONENTS` env var in `pr_check.sh`

Do you also want error message when component does not exist?


bsquizz (approved):
I have a couple small tweaks I'll make but looks good

PR #65, opened 2021-06-17
Use -s option to enable logging from pytest tests

psegedy:


bsquizz (approved):
LGTM, but you'll want to make the same changes over here too https://gitlab.cee.redhat.com/insights-qe/iqe-dockerfiles/-/blob/master/iqe-integration-tests/iqe_runner.sh

since when we fully move to running tests w/ a CJI this script here will go away.

PR #66, opened 2021-06-22
WIP: Clean up all namespaced resources in ns reconciler

bsquizz:


bsquizz:
Need to think about whether this is really more efficient since we probably wouldn't want to delete certain namespaced resources (like RoleBindings, for example). Either way, we need to maintain a list.

bsquizz:
Decided to not go with this approach for now

PR #67, opened 2021-06-22
Get artifacts from a CJI created IQE pod

BlakeHolifield:
* Get the Minio creds from env 
* Install the `mc` client
* Make a generalized pattern for any app to use
* Wait for the correct running IQE pod to start and run
* Pipe logs to main Jenkins console
* Port-forward minio server and upload artifacts

### DIFF
--- a/cicd/cji_smoke_test.sh
+++ b/cicd/cji_smoke_test.sh
@@ -1,9 +1,17 @@
 # Run smoke tests as a ClowdJobInvocation deployed by bonfire
 
 # Env vars defined by caller:
-#IQE_PLUGINS="plugin1,plugin2" -- pytest plugins to run separated by "," #IQE_MARKER_EXPRESSION="mymarker" -- pytest marker expression
 #IQE_FILTER_EXPRESSION="something AND something_else" -- pytest filter, can be "" if no filter desired
 #NAMESPACE="mynamespace" -- namespace to deploy iqe pod into, can be set by 'deploy_ephemeral_env.sh'
+
+# In order for the deploy-iqe-cji to run correctly, we must set the marker and filter to "" if they
+# are not set in pr_check.sh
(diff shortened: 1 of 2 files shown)


cicd/cji_smoke_test.sh
```diff
@@ -0,0 +1,52 @@
...
+
+CJI_NAME="$APP_NAME-smoke-tests"
+
+function kill_port_fwd {
+    echo "Caught signal, kill port forward"
+    if [ ! -z "$PORT_FORWARD_PID" ]; then kill $PORT_FORWARD_PID; fi
+}
+
+# Invoke the CJI with the exported vars from pr_check
+pod=$(bonfire deploy-iqe-cji $APP_NAME -m $IQE_MARKER_EXPRESSION -e "clowder_smoke" --cji-name $CJI_NAME -n $NAMESPACE)
```
bsquizz, line 16: We'll also want to pass `-k $IQE_FILTER_EXPRESSION`

marker or filter could potentially be undefined... I think if you default them here in the bash script to `'""'` if undefined it should work?

cicd/cji_smoke_test.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Pipe logs to background to keep them rolling in jenkins
+oc logs -n $NAMESPACE $pod -f &
+
+# Wait for the job to Complete or Fail before we try to grab artifacts
+# condition=complete does trigger when the job fails
+oc wait --timeout=$IQE_CJI_TIMEOUT --for=condition=Complete -n $NAMESPACE job/$CJI_NAME-iqe
```
bsquizz, line 23: Can you document that this env variable is needed up top in the comments

And we may want to set a default timeout if one was not set?

bsquizz (approved):

PR #68, opened 2021-06-29
Add `bonfire config edit` command for easy editing of local config

lindgrenj6:
:warning: First off: I am not a python programmer so this might be completely wrong. Roast me at your discretion!

I found myself wanting to immediately run `bonfire config edit` after running `bonfire config write-default` the first time - so I went ahead and added it here. I can't get it to quite run locally (somethign somethign `Starting path not found`), but from the looks of the `write-default` subcommand this might do the trick.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -756,6 +756,13 @@ def _cmd_write_default_config(path):
     conf.write_default_config(path)
 
 
+@config.command("edit")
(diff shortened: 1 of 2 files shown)


bsquizz (approved):
Looks good, thanks!

bsquizz:
Tested it out and it is working for me whether path is specified or not.

PR #70, opened 2021-07-01
RHCLOUD-14977 --no-remove-resources for specific components

psegedy:
- use --remove-resources/--no-remove-resources with argument e.g. `--remove-resources all --no-remove-resources vmaas --no-remove-resources something-else`
- don't allow to use `--remove-resources all --no-remove-resources all`
- `COMPONENTS_W_RESOURCES`env var for PR check to set which components should respect resources settings from clowdapp
- **it's a breaking change** - `--remove-resources/--no-remove-resources` expects 1 argument


bsquizz (approved):
Tested out a bit locally, looks good

PR #78, opened 2021-07-13
Quote marker/filter env vars

bsquizz:
Quote the marker/filter expressions since they may be multi-word expressions


bsquizz:
Handed via #79

PR #81, opened 2021-07-16
Cache image build from existing image

psegedy:
sadly, `podman` can't do such caching
```
--cache-from string    Images to utilise as potential cache sources. The build process does not currently support caching so this is a NOOP.
```

a user would set `CACHE_FROM_LATEST_IMAGE=true` to use this feature, it tries to pull an image without tag, so it assumes `:latest` tag would exist, otherwise it won't use `--cache-from`

I didn't implemented looking for latest tagged image or latest commit hash, since there might be some private repos.

### DIFF
--- a/cicd/build.sh
+++ b/cicd/build.sh
@@ -3,6 +3,7 @@
 # Env vars caller defines:
 #IMAGE="quay.io/myorg/myapp" -- docker image URI to push to
 #DOCKERFILE=Dockerfile.custom  -- dockerfile to use (optional)
+#CACHE_FROM_LATEST_IMAGE=true  -- build image from cache from latest image (optional)
 
 # Env vars set by bootstrap.sh:
 #IMAGE_TAG="abcd123" -- image tag to push to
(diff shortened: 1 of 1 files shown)


bsquizz (approved):

cicd/build.sh
```diff
@@ -41,7 +43,13 @@ if test -f /etc/redhat-release && grep -q -i "release 7" /etc/redhat-release; th
     mkdir -p "$DOCKER_CONF"
     docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
     docker --config="$DOCKER_CONF" login -u="$RH_REGISTRY_USER" -p="$RH_REGISTRY_TOKEN" registry.redhat.io
-    docker --config="$DOCKER_CONF" build -t "${IMAGE}:${IMAGE_TAG}" $APP_ROOT -f $APP_ROOT/$DOCKERFILE
+    if [ "$CACHE_FROM_LATEST_IMAGE" == "true" ]; then
+        docker --config="$DOCKER_CONF" pull "${IMAGE}" && \
+        docker --config="$DOCKER_CONF" build -t "${IMAGE}:${IMAGE_TAG}" $APP_ROOT -f $APP_ROOT/$DOCKERFILE --cache-from "${IMAGE}" || \
```
bsquizz, line 48: Can you add an 'echo' message here after the `||` to indicate that the build w/ cache failed and you are falling back to use the standard build?

bsquizz: I'll merge now and add it

PR #83, opened 2021-07-19
Fix ClowdApp name for cji

psegedy:
`APP_NAME` is the name of folder in app-sre
`COMPONENT_NAME` is the actual ClowdApp name - name of resourceTemplate in app-sre


bsquizz:
@BlakeHolifield FYI. This won't impact your rbac PR since it also has COMPONENT_NAME set.

PR #84, opened 2021-07-20
Don't look up commit hash for master/main/stable

psegedy:
Avoid gh api rate limiting by using branch name as ref for github raw file, there is no need to access gh api

However we need to look up commit hash if PR is not rebased and this can reach gh api limit again. So this is a workaround, not the actual fix for the issue.


psegedy:
nope, this is not going to work, we need commit hash to properly set image tag

PR #85, opened 2021-07-22
Fix gh rate limit

psegedy:


bonfire/utils.py
```diff
@@ -175,7 +190,8 @@ def _get_ref(self, get_ref_func):
                 else:
                     alts = ", ".join(self._alternate_refs[self.ref])
                     raise Exception(
-                        f"failed to fetch git ref {self.ref} or any of its alternates: {alts}"
+                        f"failed to fetch git ref '{self.ref}' or any of its alternates: '{alts}' "
+                        f"request failed with '{response.text}'"
```
bsquizz, line 194: If we are going to include `response.tex`t can we also list `response.status_code`

bonfire/utils.py
```diff
@@ -143,6 +145,19 @@ def _gl_certfile(self):
...
 
+    @cached_property
+    def _gh_auth_headers(self):
+        gh_token = os.getenv("GITHUB_TOKEN")
+        if gh_token:
+            return {"Authorization": f"token {gh_token}"}
+        return None
+
+    @property
+    def _jenkins_or_local(self):
```
bsquizz, line 156: I see what you are going for, IMO this might be cleaner if we just have:
```python
GH_API_URL = os.getenv('GITHUB_API_URL', "https://api.github.com")
GH_BRANCH_URL = GH_API_URL.rstrip('/') + "/repos/{org}/{repo}/git/refs/heads/{branch}"
```

Then we will just configure the app-sre Jenkins jobs to use:
GITHUB_API_URL
GITHUB_TOKEN

and then the bonfire code doesn't need to have the concept of "on jenkins" or "not on jenkins" baked into its code.

bsquizz:
Will address via #88

PR #88, opened 2021-07-23
Allow GITHUB_API_URL and GITHUB_TOKEN to be customized

bsquizz:
This is so we can make use of app-sre's github mirror utility


psegedy:
This is cleaner than my initial implementation, thank you for taking it over

PR #92, opened 2021-08-06
Support stringData for secret replacement

### LINKED ISSUE #91: import-secrets does not support json secrets with stringData key definitions
I tried importing secrets with the "stringData" definition for values, but they were ignored. Only secrets with the "data" section were honored.

I `think` the logic here https://github.com/RedHatInsights/bonfire/pull/89/files#diff-98f3050c7cb8942de6a71922230ff5a5cb2baa61c0309f51485a2c7751a7e74bR53 only checks for data before deciding on whether or not to import.

bsquizz:
Fixes #91

### DIFF
--- a/bonfire/secrets.py
+++ b/bonfire/secrets.py
@@ -50,7 +50,9 @@ def _import_secret(secret_name, secret_data):
     current_secret = get_json("secret", secret_name) or {}
 
     # avoid race conditions when running multiple processes by comparing the data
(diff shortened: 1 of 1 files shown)


abellotti:
Thanks @bsquizz for the fast turn-around 👍 

I've applied this diff to my local bonfire package and it worked like a Champ. LGTM!!

PR #93, opened 2021-08-09
Update gql to 3.0.0a6 and reset requirements

bsquizz:
We had some requirements listed in requirements.txt that we don't really need.


vidyanambiar (commented):
@bsquizz  Thank you for the quick turnaround!

PR #98, opened 2021-08-18
Fix misuse of := operator

Ichimonji10:
The `:=` operator optionally assigns a variable, and then evaluates that
variable:

```console
$ echo "${var}"

$ echo "${var:=foo}"
foo
$ echo "${var:=bar}"
foo
```

If assigning a default value to a variable, the following is idiomatic:

```console
$ : "${var:=foo}"
$ : "${var:=bar}"
$ echo "${var}"
foo
```

The idiomatic form is compatible with readonly variables:

```console
$ readonly var=foo
$ : "${var:=bar}"
$ echo "${var}"
foo
```

If assigning a default value to a variable, the following is
non-idiomatic:

```console
$ var="${var:=foo}"
$ var="${var:=bar}"
$ echo "${var}"
foo
```

The non-idiomatic form is incompatible with readonly variables, and it
wastefully performs an unnecessary assignment to boot:

```console
$ readonly var=foo
$ var="${var:=bar}"
-bash: var: readonly variable
```


Ichimonji10:
This PR is designed to address an issue I encountered while authoring https://github.com/RedHatInsights/config-manager/pull/78.

bsquizz (approved):

Ichimonji10:
In case anyone is wondering why I'm advocating for this instead of:

```bash
var="${var:-val}"
```

It's because this^ statement performs an unconditional assignment, whereas the following performs an assignment only if `var` is unset or null:

```bash
: "${var:=val}"
```

The conditional assignment allows `var` to be `readonly`:

```bash
readonly var=foo
: "${var:=bar}"
```

PR #99, opened 2021-08-18
Drop unused function `kill_port_fwd`

Ichimonji10:
```console
$ git grep kill_port_fwd
cicd/cji_smoke_test.sh:function kill_port_fwd {
cicd/deploy_ephemeral_db.sh:function kill_port_fwd {
```

### DIFF
--- a/cicd/cji_smoke_test.sh
+++ b/cicd/cji_smoke_test.sh
@@ -15,11 +15,6 @@ IQE_FILTER_EXPRESSION="${IQE_FILTER_EXPRESSION:='""'}"
 
 CJI_NAME="$COMPONENT_NAME-smoke-tests"
 
-function kill_port_fwd {
-    echo "Caught signal, kill port forward"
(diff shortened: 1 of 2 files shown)


bsquizz:
I think this function *is* supposed to be called by a trap ;) Let me look.

bsquizz:
See https://github.com/RedHatInsights/bonfire/pull/100

bsquizz:
Ok we decided since you can't re-define a trap for the same signal, to just not bother killing the port forwards at all for now. They should die when the ephemeral env is cleaned up.

Ichimonji10:
In case anyone is wondering what happens if you re-define a trap for a signal:

```bash
#!/usr/bin/env bash
# coding=utf-8

function teardown1 {
    echo "teardown1 is executing"
}

function teardown2 {
    echo "teardown2 is executing"
}

trap "teardown1" EXIT ERR SIGINT SIGTERM
trap "teardown2" EXIT ERR SIGINT SIGTERM
```

```console
$ ./test.sh 
teardown2 is executing
```

PR #100, opened 2021-08-19
Run 'kill_port_fwd' at shutdown

bsquizz:
The trap was not calling the correct function.


bsquizz:
Turns out you can't re-define a trap for the same signals twice. So we'd have to handle this in `teardown` in _common_deploy_logic.sh -- I think it is probably unnecessary to kill to port fwds anyways since the ephemeral env they are pointing to gets torn down.

PR #110, opened 2021-09-03
Patch image_tag replacement

### LINKED ISSUE #109: Image tag replacement doesn't work with image tags with non-word characters
there's an issue when the image tags to be replaced that are not just made out of word characters, such as ,  `latest-new`
In this case, the bug affects the following way:  Let's say your template has set an image tag like : 
```
your-awesome-image:latest-new
```

then the image tag replacement is set to ... say , `your-awesome-image=foobar`, so you expect the `latest-new` tag to be replaced with `foobar`.  The processor will try to replace the match for your "image" value, and then will capture whatever sequence of word characters follow a colon character. 

In this case, the replacement will (wrongly) set your image to:

```
your-awesome-image:foobar-new
```

Victoremepunto:
This aims to fix  #109 

This processes every character found up until next `"` character, which is the next "end of field character" in a JSON string.

### DIFF
--- a/bonfire/processor.py
+++ b/bonfire/processor.py
@@ -224,7 +224,7 @@ class TemplateProcessor:
         content = json.dumps(items)
         for image, image_tag in self.image_tag_overrides.items():
             # easier to just re.sub on a whole string
-            content, subs = re.subn(rf"{image}:[^\"]*", rf"{image}:{image_tag}", content)
+            content, subs = re.subn(rf"{image}:[-\w\.]+", rf"{image}:{image_tag}", content)
             if subs:
                 log.info("replaced %d occurence(s) of image tag for image '%s'", subs, image)
         return json.loads(content)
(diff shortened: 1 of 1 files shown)


bonfire/processor.py
```diff
@@ -224,7 +224,7 @@ def _sub_image_tags(self, items):
         content = json.dumps(items)
         for image, image_tag in self.image_tag_overrides.items():
             # easier to just re.sub on a whole string
-            content, subs = re.subn(rf"{image}:\w+", rf"{image}:{image_tag}", content)
+            content, subs = re.subn(rf"{image}:[^\"]*", rf"{image}:{image_tag}", content)
```
bsquizz, line 227: As we discuessed lets use `rf"{image}:[-\w\.]+"` to match any word character OR a dash OR a period -- which should be the only valid characters we would encounter in an image tag string.

bsquizz (changes requested):

bonfire/processor.py
Victoremepunto: Agree, done.

bsquizz (approved):

PR #111, opened 2021-09-07
Add deployment timeout env variable

Victoremepunto:


akofink (approved):

cicd/deploy_ephemeral_env.sh
```diff
@@ -11,5 +11,6 @@ bonfire deploy \
     --set-template-ref ${APP_NAME}/${COMPONENT_NAME}=${GIT_COMMIT} \
     --set-image-tag ${IMAGE}=${IMAGE_TAG} \
     --namespace ${NAMESPACE} \
+    --timeout ${DEPLOY_TIMEOUT:-300} \
```
akofink, line 14: @bsquizz this isn't already exposed via an env variable, is it?

bsquizz: There is not yet. But can we put this in the common deploy logic script and use the same syntax used here for setting the default value... Jeremy Audet made some cases for why using this syntax is better.

https://github.com/RedHatInsights/bonfire/blob/9668507e6b15469380f20a0322b73dcebf4a300a/cicd/_common_deploy_logic.sh#L15

Also can we add a comment up top explaining what the var does: for ex https://github.com/RedHatInsights/bonfire/blob/9668507e6b15469380f20a0322b73dcebf4a300a/cicd/_common_deploy_logic.sh#L5

Victoremepunto: @bsquizz  done, updated! 
I saw more scripts that could potentially use this variable but we only need it here, and I don't want to create any unwanted side-effect

cicd/deploy_ephemeral_env.sh
```diff
@@ -11,5 +11,6 @@ bonfire deploy \
     --set-template-ref ${APP_NAME}/${COMPONENT_NAME}=${GIT_COMMIT} \
     --set-image-tag ${IMAGE}=${IMAGE_TAG} \
     --namespace ${NAMESPACE} \
+    --timeout ${DEPLOY_TIMEOUT} \
```
akofink, line 14:
```suggestion
    ${TIMEOUT_ARG} \
```

cicd/_common_deploy_logic.sh
```diff
@@ -14,6 +15,7 @@ set -ex
 
 : ${COMPONENTS:=""}
 : ${COMPONENTS_W_RESOURCES:=""}
+: ${DEPLOY_TIMEOUT:="300"}
```
akofink, line 18: Copy the COMPONENTS/COMPONENTS_W_RESOURCES method:

```suggestion
if [ ! -z "$DEPLOY_TIMEOUT" ]; then
    export TIMEOUT_ARG=$(-t $DEPLOY_TIMEOUT)
fi
```

bsquizz: That approach is calling a func to do something a bit more complex -- I think the approach here works ok

bsquizz (approved):

PR #112, opened 2021-09-07
temporary pin down setuptools 58.x for breaking

RonnyPfannschmidt:
https://setuptools.readthedocs.io/en/latest/history.html#v58-0-2
2to3 parameters used in some dependencies start to break build pipelines


thearifismail (approved):
LGTM

PR #116, opened 2021-09-08
Initial commit to get bonfire working with ephemeral namespace operator

Jason-RH:
Hey @bsquizz this PR adds a few bonfire commands to interact with the new NamespaceReservation CRD that the namespace operator uses. Wanted to go ahead and get some eyes on it to see if you like the direction. 

Added a new "reservation" command grouping that includes:
- create
- extend
- delete
- list

There's some duplicated code, but if you like the approach I can do a bit of refactoring/cleanup. Tried to adhere to existing patterns wherever possible.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -19,10 +19,13 @@ from bonfire.openshift import (
     find_clowd_env_for_ns,
     wait_for_clowd_env_target_ns,
     wait_on_cji,
+    wait_on_reservation,
+    get_reservation_by_requester,
+    oc,
 )
 from bonfire.utils import FatalError, split_equals, find_what_depends_on
 from bonfire.local import get_local_apps
-from bonfire.processor import TemplateProcessor, process_clowd_env, process_iqe_cji
+from bonfire.processor import TemplateProcessor, process_clowd_env, process_iqe_cji, process_reservation
 from bonfire.namespaces import (
     Namespace,
     get_namespaces,
@@ -85,6 +88,12 @@ def apps():
     pass
 
 
+@main.group()
+def reservation():
+    """Perform operations related to the NamespaceReservation CRD"""
+    pass
+
+
 def _warn_if_unsafe(namespace):
     ns = Namespace(name=namespace)
     if not ns.owned_by_me and not ns.available:
@@ -93,6 +102,14 @@ def _warn_if_unsafe(namespace):
         ):
             click.echo("Aborting")
             sys.exit(0)
+
+
+def _warn_before_delete(reservation, namespace_name):
+    if not click.confirm(
+        "Deleting your reservation will also delete the associated namespace. Proceed?"
+    ):
+        click.echo("Aborting")
+        sys.exit(0)
 
 
 def _reserve_namespace(duration, retries, namespace):
@@ -1014,6 +1031,135 @@ def _cmd_apps_what_depends_on(
     apps = _get_apps_config(source, target_env, None, local_config_path)
     found = find_what_depends_on(apps, component)
     print("\n".join(found) or f"no apps depending on {component} found")
+
+
+@reservation.command("create")
+@click.option('--requester', '-r', prompt='Requester', type=str, help='Name of the user requesting a reservation')
+@click.option('--duration', '-d', prompt='Duration', type=str, default='1h', help='Duration of the reservation')
+@options(_timeout_option)
+def _create_new_reservation(requester, duration, timeout):
+    def _err_handler(err):
+        msg = f"reservation failed: {str(err)}"
+        _error(msg)
+
+    try:
+        res_config = process_reservation(requester, duration, extension=False)
+
+        log.debug("processed reservation:\n%s", res_config)
+
+        try:
+            res_name = res_config["items"][0]["metadata"]["name"]
+        except (KeyError, IndexError):
+            raise Exception("error parsing name of Reservation from processed template, check Reservation template")
+
+        apply_config(None, list_resource=res_config)
+
+        ns_name = wait_on_reservation(res_name, timeout)
+    except KeyboardInterrupt as err:
+        log.error("aborted by keyboard interrupt!")
+        _err_handler(err)
+    except TimedOutError as err:
+        log.error("hit timeout error: %s", err)
+        _err_handler(err)
+    except FatalError as err:
+        log.error("hit fatal error: %s", err)
+        _err_handler(err)
+    except Exception as err:
+        log.exception("hit unexpected error!")
+        _err_handler(err)
+    else:
+        log.info(
+            "namespace '%s' is reserved by '%s' for '%s'", ns_name, requester, duration
+        )
+        click.echo(ns_name)
+
+
+@reservation.command("extend")
+@click.option('--requester', '-r', prompt='Requester', type=str, help='Name of the user requesting an extension')
+@click.option('--duration', '-d', prompt='Duration', type=str, default='1h', help='Duration of the extension')
+def _extend_reservation(requester, duration):
+    def _err_handler(err):
+        msg = f"reservation extension failed: {str(err)}"
+        _error(msg)
+
+    try:
+        res_config = process_reservation(requester, duration, extension=True)
+
+        log.debug("processed reservation:\n%s", res_config)
+
+        apply_config(None, list_resource=res_config)
+    except KeyboardInterrupt as err:
+        log.error("aborted by keyboard interrupt!")
+        _err_handler(err)
+    except TimedOutError as err:
+        log.error("hit timeout error: %s", err)
+        _err_handler(err)
+    except FatalError as err:
+        log.error("hit fatal error: %s", err)
+        _err_handler(err)
+    except Exception as err:
+        log.exception("hit unexpected error!")
+        _err_handler(err)
+    else:
+        log.info(
+            "reservation for '%s' extended by '%s'", requester, duration
+        )
+
+
+@reservation.command("delete")
+@click.option('--requester', '-r', prompt='Requester', type=str, help='Name of the user requesting an extension')
+def _delete_reservation(requester):
+    def _err_handler(err):
+        msg = f"reservation deletion failed: {str(err)}"
+        _error(msg)
+
+    try:
+        res = get_reservation_by_requester(requester)
+        if res:
+            res_name = res["metadata"]["name"]
+            _warn_before_delete(res_name, res["status"]["namespace"])
+            log.info("deleting reservation '%s' for requester '%s'", res_name, requester)
+            oc("delete", "reservation", res_name)
+        else:
+            raise FatalError("Existing reservation for '%s' not found", requester)
+    except KeyboardInterrupt as err:
+        log.error("aborted by keyboard interrupt!")
+        _err_handler(err)
+    except TimedOutError as err:
+        log.error("hit timeout error: %s", err)
+        _err_handler(err)
+    except FatalError as err:
+        log.error("hit fatal error: %s", err)
+        _err_handler(err)
+    except Exception as err:
+        log.exception("hit unexpected error!")
+        _err_handler(err)
+    else:
+        log.info(
+            "reservation for '%s' deleted", requester
+        )
+
+
+@reservation.command("list")
+def _list_reservations():
+    def _err_handler(err):
+        msg = f"reservation listing failed: {str(err)}"
+        _error(msg)
+
+    try:
+        oc("get", "reservation")
+    except KeyboardInterrupt as err:
+        log.error("aborted by keyboard interrupt!")
+        _err_handler(err)
+    except TimedOutError as err:
+        log.error("hit timeout error: %s", err)
+        _err_handler(err)
+    except FatalError as err:
+        log.error("hit fatal error: %s", err)
+        _err_handler(err)
+    except Exception as err:
+        log.exception("hit unexpected error!")
+        _err_handler(err)
 
 
 def main_with_handler():
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -20,8 +20,10 @@ from bonfire.openshift import (
     wait_for_clowd_env_target_ns,
     wait_on_cji,
     wait_on_reservation,
-    get_reservation_by_requester,
+    get_reservation,
+    check_for_existing_reservation,
     oc,
+    whoami,
 )
 from bonfire.utils import FatalError, split_equals, find_what_depends_on
 from bonfire.local import get_local_apps
@@ -112,6 +114,15 @@ def _warn_if_unsafe(namespace):
 def _warn_before_delete(reservation, namespace_name):
     if not click.confirm(
         "Deleting your reservation will also delete the associated namespace. Proceed?"
+    ):
+        click.echo("Aborting")
+        sys.exit(0)
+
+
+def _warn_of_existing(requester):
+    if not click.confirm(
+        f"Existing reservation(s) detected for requester '{requester}'. "
+        "Do you need to reserve an additional namespace?"
     ):
         click.echo("Aborting")
         sys.exit(0)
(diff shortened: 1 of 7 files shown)


Jason-RH:
Can't merge this until https://issues.redhat.com/browse/RHCLOUD-16215 has been resolved

bonfire/bonfire.py
```diff
@@ -1016,6 +1038,170 @@ def _cmd_apps_what_depends_on(
     print("\n".join(found) or f"no apps depending on {component} found")
 
 
+@reservation.command("create")
+@click.option(
+    '--requester',
+    '-r',
+    prompt='Requester',
```
bsquizz, line 1045: Can we default to using 'oc whoami' if user does not provide the requester

bonfire/bonfire.py
```diff
@@ -1016,6 +1038,170 @@ def _cmd_apps_what_depends_on(
...
+
+    try:
+        res = get_reservation_by_requester(requester)
+        if res:
+            res_name = res["metadata"]["name"]
+            _warn_before_delete(res_name, res["status"]["namespace"])
+            log.info("deleting reservation '%s' for requester '%s'", res_name, requester)
+            oc("delete", "reservation", res_name)
+        else:
+            raise FatalError("Existing reservation for '%s' not found", requester)
```
bsquizz, line 1164: Use f-string here

bonfire/bonfire.py
```diff
@@ -1016,6 +1038,170 @@ def _cmd_apps_what_depends_on(
...
+    try:
+        res_config = process_reservation(requester, duration, extension=False)
+
+        log.debug("processed reservation:\n%s", res_config)
+
+        try:
+            res_name = res_config["items"][0]["metadata"]["name"]
+        except (KeyError, IndexError):
+            raise Exception(
+                "error parsing name of Reservation from processed template, " +
```
bsquizz, line 1072: "+" is technically not needed :)

bonfire/openshift.py
```diff
@@ -732,3 +732,33 @@ def _pod_found():
...
+        res = get_json("reservation", name=res_name)
+        try:
+            return res["status"]["namespace"]
+        except (KeyError, IndexError):
+            return False
+
+    ns_name, elapsed = wait_for(
+        _find_reservation,
+        num_sec=timeout,
+        message=f"wait for namespace to be owned by reservation '{res_name}'",
```
bsquizz, line 750: Should this say "waiting for namespace to be allocated"

bonfire/openshift.py
```diff
@@ -732,3 +732,33 @@ def _pod_found():
...
+    )
+    return ns_name
+
+
+def get_reservation_by_requester(requester):
```
bsquizz, line 755: I think we should allow for ns lookup/ns extension using:

* reservation name
* namespace name (find the reservation associated with namespace "X")

bonfire/bonfire.py
```diff
@@ -1016,6 +1038,170 @@ def _cmd_apps_what_depends_on(
...
+    type=str,
+    default='1h',
+    help='Duration of the reservation',
+)
+@options(_timeout_option)
+def _create_new_reservation(requester, duration, timeout):
+    def _err_handler(err):
+        msg = f"reservation failed: {str(err)}"
+        _error(msg)
+
```
bsquizz, line 1062: Maybe for average/"non-bot" users: check if requester already has a reservation -- if so warn them and ask "Do you really want to reserve another one?"

We should have a cli flag so that PR checks and bots don't have to answer this check

bonfire/processor.py
```diff
@@ -117,6 +117,43 @@ def process_iqe_cji(
...
+
+    with template_path.open() as fp:
+        template_data = yaml.safe_load(fp)
+
+    params = dict()
+    params["DURATION"] = duration
+    params["REQUESTER"] = requester
+
+    if extension:
+        res = get_reservation_by_requester(requester)
```
bsquizz, line 141: This part probably needs to change if the way we extend reservations changes as discussed above

bonfire/processor.py
```diff
@@ -117,6 +117,43 @@ def process_iqe_cji(
...
+        if res:
+            params["NAME"] = res["metadata"]["name"]
+        else:
+            raise FatalError("Existing reservation for '%s' not found", requester)
```
bsquizz, line 145: Use f string

bonfire/bonfire.py
Jason-RH: Added a flag (--bot/-b) that will ignore the existing reservation check

bonfire/openshift.py
Jason-RH: Changed reservation lookup to either use reservation name or attached namespace. Left out retrieving reservations via requester, and instead added some additional filtering capabilities to "list" to print out reservations via "--mine" (uses oc whoami) or "--requester".

bonfire/bonfire.py
Jason-RH: Updated this to default to using 'oc whoami' whenever requester is not present. Note: 'oc whoami' will most likely error when running locally in minikube so in that case I set the requester to 'bonfire'

bonfire/bonfire.py
```diff
@@ -1016,6 +1073,198 @@ def _cmd_apps_what_depends_on(
...
+            res_config["items"][0]["spec"]["requester"],
+            duration,
+        )
+        click.echo(ns_name)
+
+
+@reservation.command("extend")
+@click.option(
+    '--namespace',
+    '-ns',
```
bsquizz, line 1142: Can we use '-n' for the short arg here since that is used for --namespace in all other commands?

bonfire/bonfire.py
```diff
@@ -463,6 +496,30 @@ def _validate_resource_arguments(ctx, param, value):
     ),
 ]
 
+_reservation_process_options = [
+    click.option(
+        "--name",
+        "-n",
```
bsquizz, line 502: You can just remove the short arg here entirely, and require people to type out `--name` fully -- I guess the only question is -- do we expect it is going to be less common to extend a reservation by its actual name vs ns?

If we expect it to be used a lot maybe a short arg is needed... otherwise I'd say only a long arg is fine.

bonfire/bonfire.py
```diff
@@ -1016,6 +1073,198 @@ def _cmd_apps_what_depends_on(
...
+@options(_reservation_process_options)
+def _extend_reservation(name, namespace, requester, duration):
+    def _err_handler(err):
+        msg = f"reservation extension failed: {str(err)}"
+        _error(msg)
+
+    try:
+        res = get_reservation(name, namespace)
+        if res:
+            res_config = process_reservation(name, res["spec"]["requester"], duration)
```
bsquizz, line 1156: I tried to extend for an ns that does not exist and the error didn't seem to get handled cleanly:

```
❯ bonfire reservation extend -ns foo
2021-09-23 15:26:10 [    INFO] [          MainThread] Retrieving reservation
2021-09-23 15:26:11 [   ERROR] [          MainThread] hit unexpected error!
Traceback (most recent call last):
  File "/home/bsquizza/repos/bonfire/bonfire/bonfire.py", line 1156, in _extend_reservation
    res_config = process_reservation(name, res["spec"]["requester"], duration)
KeyError: 'spec'
ERROR: reservation extension failed: 'spec'
```

Not sure how we got to line 1156, would have thought `if res` would have evaluated as `False`

bonfire/bonfire.py
```diff
@@ -1016,6 +1073,198 @@ def _cmd_apps_what_depends_on(
...
+        log.exception("hit unexpected error!")
+        _err_handler(err)
+    else:
+        log.info(
+            "reservation '%s' extended by '%s'", name, duration
+        )
+
+
+@reservation.command("delete")
+@click.option(
```
bsquizz, line 1182: Should 'delete' and 'extend' basically have the same CLI args?

One could delete by: requester, name, or ns

See `_ns_reserve_options` as an example of how you can define the click options once and re-use them twice for 2 different commands

Jason-RH: Yeah I can create a `_reservation_lookup_options` for both of these commands

bonfire/bonfire.py
Jason-RH: I'm not sure. I'm fine switching to just the long arg for name though.

bonfire/bonfire.py
Jason-RH: :+1:

bonfire/bonfire.py
```diff
@@ -1137,23 +1159,21 @@ def _err_handler(err):
...
 
 @reservation.command("extend")
-@click.option(
-    '--namespace',
-    '-ns',
-    type=str,
-    default=None,
-    help='Name of the reserved namespace',
-)
+@options(_reservation_lookup_options)
```
bsquizz, line 1162: Seeing double options on this command now:
```
❯ bonfire reservation extend --help
Usage: bonfire reservation extend [OPTIONS]

Options:
  --name TEXT           Identifier for the reservation
  -r, --requester TEXT  Name of the user requesting a reservation
  -n, --namespace TEXT  Namespace for the reservation
  --name TEXT           Identifier for the reservation
  -r, --requester TEXT  Name of the user requesting a reservation
  -d, --duration TEXT   Duration of the reservation
  -h, --help            Show this message and exit.
```

bsquizz: Is it because of `@options(_reservation_process_options)` below?

bonfire/bonfire.py
```diff
@@ -1016,6 +1095,190 @@ def _cmd_apps_what_depends_on(
...
+            res_config["items"][0]["spec"]["requester"],
+            duration,
+        )
+        click.echo(ns_name)
+
+
+@reservation.command("extend")
+@options(_reservation_lookup_options)
+@options(_reservation_process_options)
+def _extend_reservation(name, namespace, requester, duration):
```
bsquizz, line 1164: I think we might want some input validation on 'duration'? I was able to pass `4` to it and the command accepted it, but that actually appears to do nothing to the reservation duration. See usage of `_validate_set_image_tag` a bit higher up in this file as an example of using a callback for parameter validation. Is there a regex we can match against for the format we're expecting for 'duration'?

```
❯ bonfire reservation list
2021-09-29 12:04:35 [    INFO] [          MainThread] running (pid 171315): oc get reservation 
2021-09-29 12:04:35 [    INFO] [          pid-171315]  |stdout| NAME                           REQUESTER   READY     NAMESPACE          EXPIRATION
2021-09-29 12:04:35 [    INFO] [          pid-171315]  |stdout| bonfire-reservation-6a60d6c5   bonfire     true      ephemeral-jqacy6   2021-09-29T20:00:41Z
❯ bonfire reservation extend -r bonfire -d 4
2021-09-29 12:04:47 [    INFO] [          MainThread] processing namespace reservation
2021-09-29 12:04:47 [    INFO] [          MainThread] running (pid 171433): oc apply -f - 
2021-09-29 12:04:48 [    INFO] [          pid-171433]  |stdout| namespacereservation.cloud.redhat.com/bonfire-reservation-6a60d6c5 configured
2021-09-29 12:04:48 [    INFO] [          MainThread] reservation 'bonfire-reservation-6a60d6c5' extended by '4'
❯ bonfire reservation list
2021-09-29 12:04:52 [    INFO] [          MainThread] running (pid 171511): oc get reservation 
2021-09-29 12:04:52 [    INFO] [          pid-171511]  |stdout| NAME                           REQUESTER   READY     NAMESPACE          EXPIRATION
2021-09-29 12:04:52 [    INFO] [          pid-171511]  |stdout| bonfire-reservation-6a60d6c5   bonfire     true      ephemeral-jqacy6   2021-09-29T20:00:41Z
```

bonfire/bonfire.py
```diff
@@ -1016,6 +1095,190 @@ def _cmd_apps_what_depends_on(
...
+        _err_handler(err)
+    else:
+        log.info(
+            "reservation '%s' extended by '%s'", res["metadata"]["name"], duration
+        )
+
+
+@reservation.command("delete")
+@options(_reservation_lookup_options)
+def _delete_reservation(name, namespace, requester):
```
bsquizz, line 1203: Hit a bug on deleting by ns, reservation name didn't get passed into the `oc delete reservation` command:

```
❯ bonfire reservation delete -n ephemeral-jqacy6
Deleting your reservation will also delete the associated namespace. Proceed? [y/N]: y
2021-09-29 12:13:06 [    INFO] [          MainThread] deleting reservation 'None'
2021-09-29 12:13:06 [    INFO] [          MainThread] running (pid 174234): oc delete reservation 
2021-09-29 12:13:06 [    INFO] [          pid-174234]  |stderr| error: resource(s) were provided, but no name, label selector, or --all flag specified
2021-09-29 12:13:06 [   ERROR] [          MainThread] hit unexpected error!
Traceback (most recent call last):
  File "/home/bsquizza/repos/bonfire/bonfire/bonfire.py", line 1213, in _delete_reservation
    oc("delete", "reservation", name)
  File "/home/bsquizza/repos/bonfire/bonfire/openshift.py", line 187, in oc
    return _exec_oc(*args, **kwargs)
  File "/home/bsquizza/repos/bonfire/bonfire/openshift.py", line 161, in _exec_oc
    raise last_err
  File "/home/bsquizza/repos/bonfire/bonfire/openshift.py", line 124, in _exec_oc
    return cmd.wait()
  File "/home/bsquizza/repos/bonfire/.venv/lib64/python3.9/site-packages/sh.py", line 879, in wait
    self.handle_command_exit_code(exit_code)
  File "/home/bsquizza/repos/bonfire/.venv/lib64/python3.9/site-packages/sh.py", line 905, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 

  RAN: /usr/bin/oc delete reservation

  STDOUT:


  STDERR:
error: resource(s) were provided, but no name, label selector, or --all flag specified

ERROR: reservation deletion failed: 

  RAN: /usr/bin/oc delete reservation

  STDOUT:


  STDERR:
error: resource(s) were provided, but no name, label selector, or --all flag specified
```

bonfire/bonfire.py
bsquizz: So after reading around a bit more, the duration is a go `time.Duration` ... maybe ensuring our 'duration' matches a regex along these lines will be good enough? https://stackoverflow.com/questions/48765574/regex-how-to-match-string-time-duration-with-optional-conditions

bsquizz: Or maybe we could handle the input validation in the CRD spec of the operator somehow instead of here in bonfire and the 'oc apply' would fail. As it stands now though, the bad duration made it through to the operator:

```
2021-09-29T12:04:58.374-0400	ERROR	controller-runtime.manager.controller.namespacereservation	Reconciler error	{"reconciler group": "cloud.redhat.com", "reconciler kind": "NamespaceReservation", "name": "bonfire-reservation-6a60d6c5", "namespace": "", "error": "time: missing unit in duration \"4\""}
```

bonfire/bonfire.py
bsquizz: Same happens trying to delete by requester name too. Deleting by reservation `--name` works though.

bonfire/bonfire.py
Jason-RH: Yes. I thought they might collapse together, but I've updated it to only use one group and pass in an extra option separately

bonfire/bonfire.py
Jason-RH: I added a short regex check in the option callback. We could probably do the validation in the operator, but the option callback is pretty easy.

bonfire/bonfire.py
Jason-RH: Yeah this is my bad. Fixed the issue in extend and forgot to apply the same fix to delete.

bonfire/bonfire.py
Jason-RH: I restricted it to XhYmZs, and added an example if an incorrect format is passed in

Jason-RH: I didn't include support for 'd' (days). We may want to add that in on the front

Jason-RH: Actually I'm just going to keep it H/M/S instead of days.

bonfire/bonfire.py
bsquizz: Still not quite right

```
❯ bonfire reservation delete -n ephemeral-qvbjz5
Deleting your reservation will also delete the associated namespace. Proceed? [y/N]: y
2021-09-30 15:35:15 [    INFO] [          MainThread] deleting reservation 'bonfire-reservation-27a266c8'
2021-09-30 15:35:15 [    INFO] [          MainThread] running (pid 460366): oc delete reservation bonfire-reservation-27a266c8 
2021-09-30 15:35:16 [    INFO] [          pid-460366]  |stdout| namespacereservation.cloud.redhat.com "bonfire-reservation-27a266c8" deleted
2021-09-30 15:35:16 [    INFO] [          MainThread] reservation 'None' deleted
```

bsquizz: Guess really its just that final log msg. Lemme look at it

PR #120, opened 2021-09-22
[WIP] Allow setting iqe requirements in CJI

psegedy:
use iqe requirements filters in pr_check using `IQE_REQUIREMENTS`, `IQE_REQUIREMENTS_PRIORITY` , and `IQE_TEST_IMPORTANCE` env vars

PR #129, opened 2021-10-13
Memory max limit to request ratio per Pod is 2

Avielyo10:
Fix:
```
Error from server (Forbidden): error when creating "STDIN": pods "iqe-tests" is forbidden: memory max limit to request ratio per Pod is 2, but provided ratio is 4
```


Avielyo10:
@kylape Seems like the CJI PR is on hold https://github.com/RedHatInsights/bonfire/pull/120

bsquizz:
Yes, CJI's are normally used now and have been for a few weeks now. You can tell if the team has `cji_smoke_test.sh` in their PR check.

But I don't mind merging this if someone still happens to be using this old script to help them out for now.

PR #140, opened 2021-10-20
Remove feature flags from eph env for now

BlakeHolifield:
Instability in the ephemeral cluster has made this issue more relevant. Let's turn it off and debug while the cluster stabilizes.


bsquizz:
Decided not to do this now since a few teams are making use of this

PR #147, opened 2021-11-03
don't quote requirements args

psegedy:
currently it results in
```
18:57:49 + iqe tests plugin vmaas --junitxml=artifacts/junit-vmaas-parallel.xml -m parallel -k '' --requirements '"VMAAS-CVES"' --requirements-priority '' --test-importance '' -n 2 --log-file=artifacts/iqe-vmaas-parallel.log
```
which won't collect any tests


psegedy:
the problem was in my pr_check

PR #152, opened 2021-11-12
Keep namespace after test run with with RELEASE_NAMESPACE=false

psegedy:
`KEEP_NAMESPACE` is unset or empty string -> release namespace
`KEEP_NAMESPACE` set to anything else -> set it to `TRUE` and keep namespace

it would be useful for debugging failing PR checks

### DIFF
--- a/cicd/_common_deploy_logic.sh
+++ b/cicd/_common_deploy_logic.sh
@@ -5,6 +5,7 @@
 #COMPONENTS="component1 component2"  # specific components to deploy (optional, default: all)
 #COMPONENTS_W_RESOURCES="component1 component2"  # components which should preserve resource settings (optional, default: none)
 #DEPLOY_TIMEOUT="300"  # bonfire deployment timeout parameter in seconds
+#KEEP_NAMESPACE=TRUE  # keep namespace after PR check for time defaulty reserverd by bonfire (optional, default: unset)
 
 # Env vars set by 'bootstrap.sh':
 #IMAGE_TAG="abcd123"  # image tag for the PR being tested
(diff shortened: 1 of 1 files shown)


cicd/_common_deploy_logic.sh
```diff
@@ -5,6 +5,7 @@
 #COMPONENTS="component1 component2"  # specific components to deploy (optional, default: all)
 #COMPONENTS_W_RESOURCES="component1 component2"  # components which should preserve resource settings (optional, default: none)
 #DEPLOY_TIMEOUT="300"  # bonfire deployment timeout parameter in seconds
+#KEEP_NAMESPACE=TRUE  # keep namespace after PR check for time defaulty reserverd by bonfire (optional, default: unset)
```
bsquizz, line 8:
```suggestion
#RELEASE_NAMESPACE="true"  # release namespace after PR check ends (default: true)
```

cicd/_common_deploy_logic.sh
```diff
@@ -67,7 +68,10 @@ function teardown {
     if [ ! -z "$NAMESPACE" ]; then
         set +e
         collect_k8s_artifacts
-        bonfire namespace release $NAMESPACE
+        if [ "${KEEP_NAMESPACE:+TRUE}" != "TRUE" ]; then
```
bsquizz, line 71:
```suggestion
        if [ "${RELEASE_NAMESPACE:-true}" != "false" ]; then
```

bsquizz (changes requested):
What if we use RELEASE_NAMESPACE env var -- by default set to true. If you do not want to release the namespace, set `RELEASE_NAMESPACE="false"` in your pr check.

psegedy:
makes sense, I've added your changes

bsquizz (approved):

PR #154, opened 2021-11-18
Bonfire v3 (switch to namespace operator)

bsquizz:


cicd/bootstrap.sh
```diff
@@ -43,7 +43,10 @@ python3 -m venv .bonfire_venv
 source .bonfire_venv/bin/activate
 
 pip install --upgrade pip 'setuptools<58' wheel
-pip install --upgrade 'crc-bonfire>=2.17.2'
+pip install --upgrade 'git+https://github.com/RedHatInsights/bonfire.git@use-ns-operator'
```
Jason-RH, line 46: I'll add a note here that this is also temporary for testing

cicd/_common_deploy_logic.sh
```diff
@@ -61,7 +61,7 @@ function collect_k8s_artifacts {
     get_oc_events
     oc get all -n $NAMESPACE -o yaml > $K8S_ARTIFACTS_DIR/oc_get_all.yaml
     oc get clowdapp -n $NAMESPACE -o yaml > $K8S_ARTIFACTS_DIR/oc_get_clowdapp.yaml
-    oc get clowdenvironment env-$NAMESPACE -o yaml > $K8S_ARTIFACTS_DIR/oc_get_clowdenvironment.yaml
+    oc get clowdenvironment $NAMESPACE -o yaml > $K8S_ARTIFACTS_DIR/oc_get_clowdenvironment.yaml
```
Jason-RH, line 64: Need to add "env-" back here

bonfire/bonfire.py
```diff
@@ -630,17 +544,18 @@ def _list_namespaces(available, mine, output):
             for ns in namespaces:
                 data[ns.name] = {
                     "reserved": ns.reserved,
-                    "ready": ns.ready,
-                    "requester": ns.requester_name,
+                    "status": ns.status,
```
bsquizz, line 547: FYI @skarekrow not sure if your iqe-tools checks this field in the JSON returned here from list namespaces ... you might need to alter the key name you're looking for when bonfire v3 is released

skarekrow: We check the `expires_in` field, but we do add the `ns.ready` field to a dict, so will note that. Thanks!

PR #156, opened 2021-11-22
Modifiable iqe_image location

dmzoneill:


bsquizz (approved):

bsquizz:
Just FYI the `smoke_test.sh` script is being phased out in favor of using `cji_smoke_test.sh` but we can still merge these changes.

PR #157, opened 2021-11-23
Click exception handling

Jason-RH:
@bsquizz attempt to refactor the exception handling to avoid duplicating so much code. This change adds a new decorator function that accepts a command string to add additional info to the error message when an exception is caught. Only using it with the namespace commands, but if you like the solution we can utilize it in the deploy commands also.


bsquizz:
Merged this into the 'use-ns-operator' branch.

PR #167, opened 2022-01-11
using unique requester when setting up ephemeral db

Jason-RH:
@bsquizz anything else required here? Using the same requester name set up in bootstrap.sh and appending "db". I didn't want to re-export the requester since env deployment also uses it


bsquizz:
Handled via https://github.com/RedHatInsights/bonfire/commit/8332aff5c5988536534567ced3b0602178ddbcbe

PR #177, opened 2022-02-03
use app name in cji smoke script

digitronik:
Most of the services use the same name for component and app folders.
in the case of ros service; we are facing a problem.

With a first look, I think it should be the app name, not the component name.
but I'm not aware of the overall implementation. I'm worried; it should not break other services. 

I think another option is to rename resourceTemplates name as. `ros` .

### DIFF
--- a/cicd/cji_smoke_test.sh
+++ b/cicd/cji_smoke_test.sh
@@ -39,7 +39,7 @@ fi
 # Invoke the CJI using the options set via env vars
 set -x
 POD=$(
-    bonfire deploy-iqe-cji $COMPONENT_NAME \
+    bonfire deploy-iqe-cji $APP_NAME \
     --marker "$IQE_MARKER_EXPRESSION" \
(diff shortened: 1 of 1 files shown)


bsquizz:
The CJI needs to be deployed using the clowdapp name which does not always match the name of the app group in app-interface. For example, the 'advisor-backend' clowdapp lives under the 'advisor' app group. When running `bonfire deploy-iqe-cji` the correct value to pass to the CJI is `advisor-backend`, not `advisor`

bsquizz:
If `APP_NAME` and `COMPONENT_NAME` are the same for your repo (which is the case for some apps) then you can just set the env vars to be the same value. For example:

https://github.com/RedHatInsights/insights-rbac/blob/master/pr_check.sh#L3-L4

bsquizz:
Closing, hopefully my above makes sense @digitronik -- let me know if you think we still need to make a change here.

PR #179, opened 2022-02-04
Ability to control how optionalDependencies are fetched

bsquizz:


bsquizz:
This adds the flag and new deps processing logic to bonfire, but does NOT yet update the pr check scripts with a new env variable. That will come in a follow-up.

PR #183, opened 2022-02-23
Fix error msg when wait_for_ready timed out but resource later goes 'ready'

bsquizz:
Sometimes the wait_for_ready function call on each `ResourceWaiter` times out. But then I believe after that sometimes the `ResourceWatcher` thread loops around and updates the status again. By the time `wait_for_ready_threaded` checks the status of all observed resources on each waiter, their status has then switched to `ready`.

This change will use a var on the ResourceWaiter to indicate if it timed out or not. We'll check that at the end instead of checking observed resources again.


bsquizz:
Tested this with an advisor deploy

PR #195, opened 2022-03-10
Adds junit-report-parser check to avoid log collection if tests passed

Victoremepunto:


Victoremepunto:
@bsquizz  I need some feedback here - wondering if an env variable to force artifact collection makes sense. Also not sure if the script is supposed to fail if no reports are found

bsquizz:
If no reports are found, that's a problem that is worth collecting logs for -- because most of the jenkins jobs do run a `junit` step after the run and if there's absolutely no `junit-*.xml` files, it is an error.

cicd/bin/check_junit_files
```diff
@@ -0,0 +1,36 @@
...
+    if report_paths:
+        junit_xml = JUnitXml()
+        for report_path in report_paths:
+            junit_xml += JUnitXml.fromfile(report_path)
+
+        if errors_or_failures_found(junit_xml):
+            sys.exit(1)
+    else:
+        print("No artifacts found")
+        # TODO: should this fail?
```
bsquizz, line 36: yes :)

cicd/bin/check_junit_files
```diff
@@ -0,0 +1,36 @@
...
+
+import logging
+import sys
+
+from glob import iglob
+from junitparser import JUnitXml
+
+if __name__ == "__main__":
+
+    def errors_or_failures_found(junit_report):
```
bsquizz, line 11: You could just define this up above outside of this scope

cicd/_common_deploy_logic.sh
```diff
@@ -74,7 +74,14 @@ function teardown {
     for ns in ${UNIQUE_NAMESPACES[@]}; do
         echo "Running teardown for ns: $ns"
         set +e
-        collect_k8s_artifacts $ns
+
+        if check_junit_files "${ARTIFACTS_DIR}/junit-*.xml"; then
```
bsquizz, line 78: We should also check if we arrived here due to an ERR signal and not a normal/successful EXIT signal

If we arrived here due to an error, we should also collect logs during teardown.

cicd/bin/check_junit_files
Victoremepunto: Done - if the glob expression does not expand into any file it will exit with error

cicd/bin/check_junit_files
Victoremepunto: yeah I forgot to move it :D - done. thanks!

cicd/_common_deploy_logic.sh
Victoremepunto: I did the proxy-trap function thing we discussed - I now check if we arrived there by anything different than a EXIT then it'll just collect artifacts. Check the current implementation and LMK what you think

Victoremepunto:
@bsquizz  ready for re-review - I also need some feedback about the expected output but I don't want to needlessly complicate this . 

- Currently it stops on the first error, failure it founds and returns 1 
- It does not show any message or rationale about what happened, just exit code 0 if files where found and no errors and failures spotted, or 1 if errors occurred (so far I handle , existing errors or failures in a report, or one of the matching files from the glob expression not being parsed by the JUnit report library. Do *note that* even one file slips in that's not a Junit report it will fail for this reason and 
- The script also could list the files matched by the glob expression, and could also show which files  made him to stop, or even keep looking through the rest (maybe not that useful and/or kills the purpose of this script).

cicd/_common_deploy_logic.sh
bsquizz: The proxy trap is a cool solution

cicd/_common_deploy_logic.sh
```diff
@@ -90,14 +92,14 @@ function teardown {
         echo "Running teardown for ns: $ns"
         set +e
 
-        if [ "$CAPTURED_SIGNAL" == "EXIT" ] && check_junit_files "${ARTIFACTS_DIR}/junit-*.xml"; then
+        if [ "$ALWAYS_COLLECT_LOGS" != "true" ] && [ "$CAPTURED_SIGNAL" == "EXIT" ] && check_junit_files "${ARTIFACTS_DIR}/junit-*.xml"; then
             echo "No errors or failures detected on JUnit reports, skipping K8s artifacts collection"
         else
-            echo "Errors or failures detected, collecting K8s artifacts"
+            [ "$ALWAYS_COLLECT_LOGS" != "true" ] && echo "Errors or failures detected, collecting K8s artifacts"
```
Victoremepunto, line 98: nitpick: the message is a bit confusing since you're actually forcing the log collection so not necessarily caused by an error or failure.  I thought of implement this with a small `if` clause for different messaging. Still splitting hairs so your call

Victoremepunto: LOL - I don't know how to read - ignore me.

PR #201, opened 2022-03-31
Add unit tests for Bonfire.py

maknop:


maknop:
Added tests for click options to get feedback on the initial approach.

bsquizz:
Thanks, FYI I think I will be merging 'dependency_changes' into the main branch tomorrow so we'll probably edit the branch this is being merged into after that.

maknop:
Sounds great!

maknop:
Added some more tests.

tests/test_bonfire.py
```diff
@@ -0,0 +1,228 @@
...
+    namespace_4.name = "namespace-4"
+    namespace_4.requester = "user-2"
+    namespace_4.expires_in = "1h"
+
+    all_namespaces.append(namespace_1)
+    all_namespaces.append(namespace_2)
+    all_namespaces.append(namespace_3)
+    all_namespaces.append(namespace_4)
+
+    all_reservations.append(namespace_1)
```
bsquizz, line 157: Similar comment to what I have about `mine=True` filter below as well.

I am not sure if this is truly asserting the behavior to check that the 'available=True' filtering is working in the get_namespaces function.

I think the more ideal thing to do would be to mock these function calls out:
https://github.com/RedHatInsights/bonfire/blob/master/bonfire/namespaces.py#L204-L206

Mock get_all_namespaces to return the 4 fake namespaces you're creating up top. Two of those fake namespaces should be set up such that their "available" function returns True: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/namespaces.py#L120

And in that way, we'd be ensuring that get_namespaces(available=True) is only returning namespaces which are ready + not reserved

tests/test_utils.py
```diff
@@ -0,0 +1,95 @@
...
+        (
+            ["t1 = test1", "t2 = test2"],
+        ),
+        (
+            ["t1 test1", "t2 test2"],
+        ),
+    ]
+)
+def test_split_equals_fail(list_of_str: list):
+    with pytest.raises(Exception):
```
bsquizz, line 42: We're expecting a `ValueError` specifically here right? Can we assert for that?

tests/test_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+@pytest.mark.parametrize(
+    "list_of_str, expected",
+    [
+        (
+            ["t1=test1", "t2=test2"],
+            {"t1": "test1", "t2": "test2"}
+        ),
+        (
+            ["company=redhat", "team=Dev Prod"],
```
bsquizz, line 19: Hmmm actually, I don't think we should allow whitespace :laughing: you may have identified a tweak we should make to the regex here

tests/test_utils.py
```diff
@@ -0,0 +1,95 @@
...
+        ),
+        (
+            "1h20s", "1h20s"
+        ),
+        (
+            "2h15m32s", "2h15m32s"
+        ),
+    ]
+)
+def test_validate_time_string(time: str, expected: str):
```
bsquizz, line 63: Could we add a test to assert that the validate FAILS and raises a ValueError for the 3 invalid cases written in that func?
* `"invalid format for duration '{time}', expecting h/m/s string. Ex: '1h30m'"
* `"invalid duration '{time}', must be less than 14 days"`
* `"invalid duration '{time}', must be more than 30 mins"`

bsquizz (changes requested):
Thanks Matt, good start... I think we could tweak things a bit to assert a little "lower level" in the code in some areas. We can chat more on Monday if you need more clarification.

Also can you add the new package names (`mock`, `pytest`, `pytest-mock`) to the 'test' extras in setup.cfg here:
https://github.com/RedHatInsights/bonfire/blob/master/setup.cfg#L41-L43

maknop:
> Thanks Matt, good start... I think we could tweak things a bit to assert a little "lower level" in the code in some areas. We can chat more on Monday if you need more clarification.
> 
> Also can you add the new package names (`mock`, `pytest`, `pytest-mock`) to the 'test' extras in setup.cfg here: https://github.com/RedHatInsights/bonfire/blob/master/setup.cfg#L41-L43

Thank you for the feedback. I'll start working through it. Much appreciated!

tests/test_bonfire.py
```diff
@@ -0,0 +1,228 @@
...
+    mocker.patch("bonfire.openshift.check_for_existing_reservation", return_value=True)
+    mocker.patch("bonfire.openshift.parse_restype", return_value="")
+    mocker.patch("bonfire.openshift.get_json", return_value="")
+    mocker.patch("bonfire.openshift.get_all_reservations", return_value="")
+    mocker.patch("bonfire.bonfire.reserve_namespace", return_value=ns)
+
+    runner = CliRunner()
+    result = runner.invoke(bonfire.namespace, ["reserve", "--requester", user])
+
+    assert result.output.rstrip() == expected
```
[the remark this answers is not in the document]
maknop: Went ahead and mocked the `process_reservation` function. Also realized I was mocking way more than I needed to, if I understand correctly.

maknop:
Added mock, pytest, pytest-mock to the 'test' extras in setup.cfg.

tests/test_bonfire.py
```diff
@@ -0,0 +1,228 @@
...
+
+    mocker.patch("bonfire.openshift.get_all_reservations", return_value="")
+    mocker.patch("bonfire.bonfire.has_ns_operator", return_value=True)
+    mocker.patch("bonfire.openshift.get_api_resources", return_value=[])
+    mocker.patch("bonfire.bonfire.get_namespaces", return_value=all_namespaces)
+
+    runner = CliRunner()
+    result = runner.invoke(bonfire.namespace, ["list"])
+
+    assert "namespace-1" in result.output
```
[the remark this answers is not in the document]
maknop: Good point. That would show for a much better test. Went ahead and implemented a new solution. Let me know what you think.

tests/test_bonfire.py
```diff
@@ -0,0 +1,228 @@
...
+    all_reservations.append(namespace_1)
+    all_reservations.append(namespace_3)
+
+    mocker.patch("bonfire.namespaces.get_all_reservations", return_value=all_reservations)
+    mocker.patch("bonfire.bonfire.has_ns_operator", return_value=True)
+    mocker.patch("bonfire.openshift.get_api_resources", return_value=[])
+    mocker.patch("bonfire.bonfire.get_namespaces", return_value=all_reservations)
+
+    runner = CliRunner()
+    result = runner.invoke(bonfire.namespace, ["list", "--mine"])
```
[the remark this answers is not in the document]
maknop: Yeah, that was an accident. Went ahead and updated it. Thank you.

tests/test_utils.py
maknop: Haha, that's funny! I'll fix the test.

tests/test_bonfire.py
```diff
@@ -0,0 +1,148 @@
...
+from click.testing import CliRunner
+import json
+
+from bonfire import bonfire
+
+
+namespace_data = {}
+reservation_data = {}
+
+with open("data/namespace_data.json", 'r') as namespace_data_file:
```
bsquizz, line 11: Let's load these files in a module-scoped fixture instead.

Also let's define a data path dynamically... if I run `pytest .` in the root dir, the tests are unable to locate `data/namespace_data.json`

Something along these lines should work:
```python
from pathlib import Path

DATA_PATH = Path(__file__).parent.joinpath("data")

@pytest.fixture(scope="module")
def namespace_list():
    with open(DATA_PATH.joinpath("namespace_data.json")) as namespace_data_file:
        return json.load(namespace_data_file)["items"]
```

bsquizz (changes requested):

PR #210, opened 2022-04-25
Rhcloud 18846 use new labels annotations

maknop:
- Switched "status" label to "env-status"
- get_all_namespaces() function filters by newly created label: "operator-ns"
- Added property for the Namespaces class in namespaces.py to get label.
- Removed "operator-ns" annotation.

### DIFF
--- a/bonfire/namespaces.py
+++ b/bonfire/namespaces.py
@@ -73,6 +73,10 @@ class Namespace:
         return self._data.get("metadata", "").get("annotations", {})
 
     @property
+    def labels(self):
+        return self._data.get("metadata", "").get("labels", {})
+
+    @property
     def status(self):
         return self.annotations.get("env-status", "false")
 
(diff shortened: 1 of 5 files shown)


bonfire/namespaces.py
```diff
@@ -185,8 +192,8 @@ def clowdapps(self):
         managed = len(self._clowdapps)
         ready = 0
         for app in self._clowdapps:
-            if "status" in app:
-                deployments = app["status"]["deployments"]
+            if "env-status" in app:
```
Jason-RH, line 195: This should be switched back to "status"

bonfire/openshift.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
Jason-RH, line 1: is this needed?

Jason-RH (approved):

PR #213, opened 2022-05-10
replace spaces with dashes for the gitlab comment

lina-is-here:
If the repo name contains spaces, the link to post a comment becomes broken. Spaces are replaced with dashes.
Tested here https://url.corp.redhat.com/cf5d9b8

### DIFF
--- a/cicd/post_test_results.sh
+++ b/cicd/post_test_results.sh
@@ -50,7 +50,7 @@ then
       -X POST \
       -H "PRIVATE-TOKEN: ${GITLAB_TOKEN_IQE_BOT}" \
       -H "Content-Type: application/json; charset=utf-8" \
-      ${GITLAB_HOST_IQE_BOT}/api/v4/projects/${gitlabTargetNamespace}%2F${gitlabTargetRepoName}/merge_requests/${gitlabMergeRequestIid}/notes \
+      ${GITLAB_HOST_IQE_BOT}/api/v4/projects/${gitlabMergeRequestTargetProjectId}/merge_requests/${gitlabMergeRequestIid}/notes \
       -d "{\"body\":\"$message\"}" -v
     set -e
   fi


cicd/post_test_results.sh
```diff
@@ -41,6 +41,8 @@ then
     done
 
     message="${message}\n${urls}"
+    # replace spaces with dashes
+    gitlabTargetRepoName="${gitlabTargetRepoName// /-}"
```
Victoremepunto, line 45: I'd recommend against manipulating external variables, this is set by the [Jenkins Gitlab plugin](https://github.com/jenkinsci/gitlab-plugin/blob/8480f1c772315a112b39df6ee099ffe0d9300810/README.md#defined-variables) , I'd also check if it's defined. 

How's this a problem BTW ? the URLs should be valid as provided in that variable.

Victoremepunto: Sorry I misread, I thought it was the Repo's URL - not name.

I was just reviewing [Gitlab API docs for adding merge-request-notes](https://docs.gitlab.com/ee/api/notes.html#create-new-merge-request-note) where it suggests the Project's ID could be used instead of the URL encoded version of the namespace path. 

If that works, then 

``` ${GITLAB_HOST_IQE_BOT}/api/v4/projects/${gitlabTargetNamespace}%2F${gitlabTargetRepoName}/merge_requests/${gitlabMergeRequestIid}/notes
```

Could be transformed into

```
/projects/:id/merge_requests/:merge_request_iid/notes
${GITLAB_HOST_IQE_BOT}/api/v4/projects/${gitlabMergeRequestTargetProjectId}/merge_requests/${gitlabMergeRequestIid}/notes
```

Victoremepunto: Okay, I tested and it works, see POC here https://github.com/Victoremepunto/bonfire/blob/test-gitlab-id/cicd/post_test_results.sh#L50

lina-is-here: perfect! I tested it too and it works :) don't know how I missed that

bsquizz (approved):

PR #215, opened 2022-05-12
Rhcloud 19331 add pool type flag

maknop:
Added **--pool-type** flag and updated reservation template. Just wanted to ensure I was headed in the correct direction.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -208,6 +208,12 @@ _ns_reserve_options = [
         help="Duration of the reservation",
         callback=_validate_reservation_duration,
     ),
+    click.option(
+        "--pool-type",
+        type=str,
+        default="default",
+        help="Specifies the pool type name. Either default or minimal",
+    ),
     _local_option,
 ]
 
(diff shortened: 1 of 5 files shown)


bonfire/bonfire.py
```diff
@@ -208,6 +208,13 @@ def _validate_reservation_duration(ctx, param, value):
         help="Duration of the reservation",
         callback=_validate_reservation_duration,
     ),
+    click.option(
+        "--pool-type",
+        type=str,
```
bsquizz, line 213: If there's only a limited set of valid choices ('default' or 'minimal') -- check out the click 'choice' type: https://click.palletsprojects.com/en/8.1.x/options/#choice-options

bonfire/resources/reservation-template.yaml
```diff
@@ -13,6 +13,7 @@ objects:
   spec:
     duration: ${DURATION}
     requester: ${REQUESTER}
+    pool-type: ${POOLTYPE}
```
bsquizz, line 16: I suspect we are going to ask that the key name here be renamed to `poolType` over in https://github.com/RedHatInsights/ephemeral-namespace-operator/pull/65/files#diff-14d05dc348d738e3ff5d3d6a0ca563b6616adb212c02c12b6628be117d81e081R30 -- so you'll probably want to update it here too :)

bsquizz (changes requested):

bonfire/resources/reservation-template.yaml
maknop: Good point. Going to shorten it to just "pool" instead.

bonfire/bonfire.py
maknop: Cool, just added this.

PR #221, opened 2022-05-17
Rhcloud 19304 unit tests

maknop:
Added two new unit tests to the test_bonfire.py file for the following commands: `bonfire namespace list --output` and `bonfire namespace reserve --timeout`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
-}
--- a/tests/test_bonfire.py
+++ b/tests/test_bonfire.py
@@ -180,3 +180,75 @@ def test_ns_list_option_mine(
     assert " ".join(["namespace-3", "false", "ready", "none"]) not in actual
     assert " ".join(["namespace-4", "false", "ready", "none"]) not in actual
     assert " ".join(["namespace-5", "true", "false", "none", "user-5"]) not in actual
+
+
+@pytest.mark.parametrize(
+    "namespace, data",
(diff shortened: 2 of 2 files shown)


tests/test_bonfire.py
```diff
@@ -164,3 +165,79 @@ def test_ns_list_option_mine(mocker, caplog, namespace_list: list, reservation_l
...
+    runner = CliRunner()
+    result = runner.invoke(bonfire.namespace, ["list", "--output", "json"])
+
+    actual_ns_1 = json.loads(result.output).get("namespace-1")
+    actual_ns_2 = json.loads(result.output).get("namespace-2")
+    actual_ns_3 = json.loads(result.output).get("namespace-3")
+    actual_ns_4 = json.loads(result.output).get("namespace-4")
+    actual_ns_5 = json.loads(result.output).get("namespace-5")
+
+    assert {
```
bsquizz, line 195: I might be wrong, but I think these asserts are just comparing length?

What about something more along the lines of:
```
d1 = {"hello": "hello", "hi": "hi", "bye": "bye"}

d2 = {"hi": "hi"}

assert all([item in d1.items() for item in d2.items()])
```

bsquizz (changes requested):

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "python.testing.unittestEnabled": false,
```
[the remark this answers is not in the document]
maknop: Oh, definitely.

tests/test_bonfire.py
bsquizz: Looks like `.items()` might be a set-like object based on what I read in https://docs.python.org/3/library/stdtypes.html#dict-views ... But its still kind of unclear. I think the above is probably more explicit and less confusing.

maknop: Just updated with this method. One thing I had to do was delete the last key/value that was returned from click because it is the expires_in data that will have a changing value each time I run the test. Once I was able to do that I could test that the other three key/value pairs were present.

bsquizz (approved):

PR #227, opened 2022-06-15
Rhcloud 19820 add pool type column

maknop:
Created a column that displays the pool type. Fixed tests that check for pool type in the results.

### DIFF
--- a/tests/test_bonfire.py
+++ b/tests/test_bonfire.py
@@ -118,7 +118,7 @@ def test_ns_list_option(mocker, caplog, namespace_list: list, reservation_list: 
 
     actual = " ".join(result.output.split())
 
-    assert " ".join(["namespace-1", "true", "false", "none", "user-1", "default"]) in actual
+    assert " ".join(["namespace-1", "true", "false", "none", "user-1", "minimal"]) in actual
     assert " ".join(["namespace-2", "true", "false", "none", "user-2", "default"]) in actual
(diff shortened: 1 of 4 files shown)


psav (commented):
I'd love to merge this - but I think there are too many unrelated changes here. I'm not sure which formatter you are using but it looks like the line length is set too low and we're seeing a bunch of changes that I don't think are necessary.

maknop:
I was using black for python formatting. I totally agree- the line length is rather short. I fixed all those changes so it should be cleaner now :)

tests/test_bonfire.py
```diff
@@ -164,18 +164,18 @@ def test_ns_list_option_mine(mocker, caplog, namespace_list: list, reservation_l
...
+    assert " ".join(["ns-4", "false", "ready", "none"]) not in actual
+    assert " ".join(["ns-5", "true", "false", "none", "user-5"]) not in actual
 
 
 def test_ns_list_flag_output(
     mocker,
     caplog,
     namespace_list: list,
-    reservation_list: list,
+    reservation_list: list
```
psav, line 178: Did this come in as a part of a format? We shouldn't change this.

psav (changes requested):
Small change

tests/test_bonfire.py
maknop: Oh no it didn't sorry. I'll fix that.

psav (approved):
Approved

PR #229, opened 2022-06-20
unify the message for GH and GL

lina-is-here:
Works for GH: https://github.com/RedHatInsights/drift-backend/pull/340
For some reason pr_check didn't run in advisor repo yet, so not tested there


lina-is-here:
upd: verified in advisor repo as well, so it's ready for review @bsquizz

bsquizz (approved):

PR #233, opened 2022-07-06
[WIP] RHCLOUD-20300 CICD init

BlakeHolifield:
* Creates a `pr_check.sh` and `build_deploy.sh` based on app name and type of project
* Has room to support frontend and backend apps
* Uses `jinja2` to generate `pr_check` and `build_deploy` files

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ The `bonfire process-env` command can be used to print a processed ClowdEnvironm
 The `bonfire deploy-env` command can be used as a helpful "1-liner" command to apply a ClowdEnvironment
 configuration into a cluster and wait for environment resources to come up successfully.
 
+The `bonfire cicd` command can be used to setup ci within your project accoriding to app-sre standards.
+
+
 Use `bonfire --help` to get a list of other commands you may find useful.
 
 ### Loading an app's ephemeral config from app-interface (default behavior)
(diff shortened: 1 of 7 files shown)


bonfire/cicd.py
```diff
@@ -0,0 +1,106 @@
...
+        os.chmod(os.path.join(os.getcwd(), target_file), 0o755)
+
+    def process_template_line(self, line):
+        if "%" in line:
+            processed = self.process_inline_var(line)
+            return processed
+        else:
+            return line
+
+    def process_inline_var(self, line):
```
bsquizz, line 76: I am thinking it would be better to use jinja2 templating instead of writing your own custom rendering/variable substitution logic. That'll also future-proof this whole thing in case you want to do more complex things in future. It's pretty simple. https://realpython.com/primer-on-jinja-templating/#quick-examples

Thoughts?

bsquizz (changes requested):

bonfire/cicd.py
BlakeHolifield: So I was initially going to use jinja, but didn't want to add another dependency to the project. I'm certainly happy to use it if that's fine with you.

bsquizz: Yep no worries on adding another dependency

README.md
```diff
@@ -37,6 +37,9 @@ The `bonfire process-env` command can be used to print a processed ClowdEnvironm
 The `bonfire deploy-env` command can be used as a helpful "1-liner" command to apply a ClowdEnvironment
 configuration into a cluster and wait for environment resources to come up successfully.
 
+The `bonfire cicd` command can be used to setup ci within your project accoriding to app-sre standards.
```
gburges, line 40: alias accoriding='according'

PR #235, opened 2022-07-13
Log to include the URL pointing to the reserved ns in OpenShift

skateman:
When running CI and I'm in need to check the deployment, having this URL can save a lot of time as links in the jenkins output are clickable.


bsquizz:
I agree this would be useful, the approach needs to be a little different because:
1. app interface base URL is actually the graphQL API URL, not the OpenShift console URL
2. we do not always use bonfire to deploy to OpenShift clusters so we have to account for the fact that sometimes a cluster URL may not exist

See https://github.com/RedHatInsights/bonfire/pull/236

PR #239, opened 2022-08-04
Add cypress deploy support

dmzoneill:
- Opening PR to start discussion on this
 - Obviously this is duplicates code from cji stubs, with the exception of the yaml file
 -


bsquizz:
Trying to get some clarification here -- I am confused because the cypress CJI YAML template looks the same as the IQE CJI template and the code to handle the process/deployment of the CJI looks the same -- but I might be missing something. What is different between a cypress CJI deploy and an IQE CJI deploy that the code in this PR is taking care of?

bsquizz:
We're moving the cicd scripts into a separate repo, so we can open an MR for new changes there in future.

PR #244, opened 2022-08-18
Add a secondary tag for PR/MR image builds

mshriver:
Within build.sh, construct `IMAGE_TAG_OPTS`
- In the case of a PR, use `pr-123-SHA` and `pr-123-latest` tags
- Docker push with --all-tags
- podman push twice

https://issues.redhat.com/browse/RHCLOUD-20838

Tested alongside this repo:
https://github.com/content-services/content-sources-backend/pull/82/files

Pipeline run:
https://ci.int.devshift.net/job/content-services-content-sources-backend-pr-check/82/console 

Resulting images:
image

### DIFF
--- a/cicd/bootstrap.sh
+++ b/cicd/bootstrap.sh
@@ -68,9 +68,7 @@ pip install --upgrade 'crc-bonfire>=4.4.0'
 
 # clone repo to download cicd scripts
 rm -fr $BONFIRE_ROOT
-# DONOTMERGE testing in pr_check against 
-# https://github.com/content-services/content-sources-backend/pull/82/files
-git clone --branch add-pr-latest-tag https://github.com/mshriver/bonfire.git $BONFIRE_ROOT
+git clone --branch master https://github.com/RedHatInsights/bonfire.git $BONFIRE_ROOT
 
 # Do a docker login to ensure our later 'docker pull' calls have an auth file created
 source ${CICD_ROOT}/_common_container_logic.sh
--- a/cicd/bootstrap.sh
+++ b/cicd/bootstrap.sh
@@ -68,7 +68,9 @@ pip install --upgrade 'crc-bonfire>=4.4.0'
 
 # clone repo to download cicd scripts
 rm -fr $BONFIRE_ROOT
-git clone --branch master https://github.com/RedHatInsights/bonfire.git $BONFIRE_ROOT
+# DONOTMERGE testing in pr_check against 
+# https://github.com/content-services/content-sources-backend/pull/82/files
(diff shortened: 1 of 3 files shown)


bsquizz (changes requested):
My suggestion to make this simpler: Just always push 2 tags and don't worry about defining IMAGE_TAG_SECONDARY or conditionally setting the IMAGE_TAG_OPTS

If the tag name starts with 'pr-', just run a 'sed' or something on the string to replace `pr-*-$IMAGE_TAG` with `pr-*-latest` and push that tag at the same time you push the other one.

mshriver:
Don't think any teams will have a desire to set the var themselves for pr_checks?  Definitely up for simplifying it if that's not a use case you think we need to cover here.

Can always make it better once its converted to python too :wink:

mshriver:
Suggestion mostly applied. Due to need to run multiple docker/podman commands to tag, I kept the use of `IMAGE_TAG_OPTS` but simplified the secondary tag injection.

cicd/bootstrap.sh
```diff
@@ -68,7 +68,9 @@ pip install --upgrade 'crc-bonfire>=4.4.0'
 
 # clone repo to download cicd scripts
 rm -fr $BONFIRE_ROOT
-git clone --branch master https://github.com/RedHatInsights/bonfire.git $BONFIRE_ROOT
+# DONOTMERGE testing in pr_check against
```
mshriver, line 71: Placeholder comment, this needs removed before merge.

cicd/build.sh
```diff
@@ -40,28 +45,34 @@ function docker_build {
...
     else
         set -x
-        docker build -t "${IMAGE}:${IMAGE_TAG}" $APP_ROOT -f $APP_ROOT/$DOCKERFILE
+        docker build $IMAGE_TAG_OPTS $APP_ROOT -f $APP_ROOT/$DOCKERFILE
         set +x
     fi
     set -x
-    docker push "${IMAGE}:${IMAGE_TAG}"
+
+    docker push --all-tags "${IMAGE}:${IMAGE_TAG}"
```
bsquizz, line 63: Question here -- does  `docker push --all-tags <image>:<image_tag>` push all tags that are ONLY the same as `image_tag` ? The only time I have seen it used is with syntax: `docker push --all-tags <image>` which pushes up ALL tags for the image.

If we aren't sure, should we just change this to use the same logic as the podman push below (using the `if`?)

bsquizz (changes requested):
1 question about `docker push --all-tags`

cicd/build.sh
mshriver: All tags for the image - I might be making a bad assumption here that the build machine would only have relevant tags for the image.

bsquizz (approved):

PR #246, opened 2022-09-15
Add dash y option to namespace reserve

psav:


bonfire/bonfire.py
```diff
@@ -218,6 +218,12 @@ def _validate_reservation_duration(ctx, param, value):
         show_default=True,
         help="Specifies the pool type name",
     ),
+    click.option(
+        "-y", "--no-prompt",
```
Victoremepunto, line 222: First - I'm surprised this hasn't arisen before. 
Second - this needs to be `-f` - I think.

Victoremepunto: yeah totally - `-f` and `--force` . `-y` paired with `--no-prompt` has a high "what was that option again" factor.

Victoremepunto (changes requested):

bsquizz (approved):
LGTM & tested manually

PR #247, opened 2022-09-23
Give a url at the end of deployment

gburges:


bonfire/bonfire.py
```diff
@@ -1070,7 +1070,10 @@ def _err_handler(err):
         _err_handler(err)
     else:
         log.info("successfully deployed to namespace '%s'", ns)
-        click.echo(ns)
+        url = namespace.get_console_url()
+        if url:
+            ns_url = f"{url}/k8s/cluster/projects/{ns}"
+        click.echo(ns_url)
```
bsquizz, line 1076: This will break some of the scripts that expect stdout to be a namespace, for ex scripts that use: ```NAMESPACE=$(bonfire deploy ...)```

If you log the ns URL instead, that would be better. A PR that does this was merged previously: https://github.com/RedHatInsights/bonfire/pull/236/ -- do you just want the log message to appear at the bottom instead?

bsquizz (changes requested):

bonfire/bonfire.py
gburges: yeah i'd like a clickable url at the end of the whole deploy process. The current one gives me a link when the namespace gets reserved, so i mostly copypasta that code :)

bsquizz (approved):

PR #249, opened 2022-10-06
Switch use 'operator-ns' label with 'pool'

maknop:
We used to identify namespaces owned by the ephemeral-namespace-operator by checking for the label `operator-ns`. Now that we have the `pool` label, we can also check all namespaces owned by the ENO by ensuring that these namespaces contain the `pool` label. This MR just removes the use of `operator-ns` and replaces it with `pool`.

### DIFF
--- a/bonfire/namespaces.py
+++ b/bonfire/namespaces.py
@@ -84,17 +84,6 @@ class Namespace:
         return self.annotations.get("reserved", "false") == "true"
 
     @property
-    def operator_ns(self):
-        return self.labels.get("operator-ns", "false") == "true"
-
-    @property
-    def is_reservable(self):
-        """
-        Check whether a namespace was created by the namespace operator.
-        """
-        return self.operator_ns
-
-    @property
     def pool_type(self):
         return self.labels.get("pool", "false")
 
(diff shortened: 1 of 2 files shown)


bonfire/namespaces.py
```diff
@@ -83,17 +83,6 @@ def status(self):
     def reserved(self):
         return self.annotations.get("reserved", "false") == "true"
 
-    @property
```
bsquizz, line 86: Could we:
1. Edit the 'pool_type' function in this class and have it return `self.labels.get("pool")`
2. Move the `def is_reservable` function to be defined below the `def pool_type` function (I just like to have functions that are used in other functions defined "higher up" in the file when possible for readability)
2. Keep the `is_reservable` property and have it return `self.pool_type is not None`

bonfire/namespaces.py
```diff
@@ -239,7 +228,7 @@ def get_namespaces(available=False, mine=False):
     ephemeral_namespaces = []
     for ns_kwargs in all_ns_kwargs:
         ns = Namespace(**ns_kwargs)
-        if not ns.is_reservable:
```
bsquizz, line 242: Keep is_reservable

bsquizz:
@maknop do we still want this to go in?

PR #251, opened 2022-10-07
Adding UI changes for namespace limiting

maknop:
This card adds UI enhancements that are directly related to adding limiting to namespace pools when `SizeLimit` is specified. This way users will know why they aren't able to reserve a namespace if the pool size limit has been reached. 

Here is an example of the output when this occurs:
```
2022-10-12 09:43:48 [    INFO] [          MainThread] Checking for available namespaces to reserve.
2022-10-12 09:43:49 [    INFO] [          MainThread] Pool size limit is defined as 3 in 'minimal' pool
ERROR: maximum number of namespaces for pool `minimal` (limit: 3) have been reserved

```

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -682,10 +682,12 @@ def _cmd_namespace_reserve(name, requester, duration, pool, timeout, local, forc
     log.info("Checking for available namespaces to reserve.")
 
     if pool_size_limit := get_pool_size_limit(pool):
+        print(pool_size_limit)
+        print(get_reserved_namespace_quantity(pool))
         log.info(f"Pool size limit is defined as {pool_size_limit} in '{pool}' pool")
         if pool_size_limit > 0 and get_reserved_namespace_quantity(pool) >= pool_size_limit:
             _error(f"maximum number of namespaces for pool `{pool}` (limit: {pool_size_limit})"
-                   "have been reserved")
+                   " have been reserved")
 
     log.info("Attempting to reserve a namespace...")
     if not has_ns_operator():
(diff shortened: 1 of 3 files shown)


bonfire/bonfire.py
```diff
@@ -670,6 +673,20 @@ def _list_namespaces(available, mine, output):
 @click_exception_wrapper("namespace reserve")
 def _cmd_namespace_reserve(name, requester, duration, pool, timeout, local):
     """Reserve an ephemeral namespace"""
+    log.info("Checking for available namespaces to reserve.")
+    if pool_size_limit := get_pool_size_limit(pool):
+        log.info(f"Pool size limit is defined as {pool_size_limit} in '{pool}' pool")
+
+        if get_reserved_namespace_quantity(pool) == get_pool_size_limit(pool):
+            log.info(f"""Namespace max reached for '{pool}' pool
```
bsquizz, line 681: Use the `_error()` function here instead of `log.info` / `exit`

bonfire/bonfire.py
```diff
@@ -670,6 +673,20 @@ def _list_namespaces(available, mine, output):
...
+
+                Max number of namespaces for '{pool}' pool have been reserved. We apologize for
```
bsquizz, line 683: Can we just keep the error message short, e.g.:
_error(f"maximum number of namespaces for pool `{pool}` (limit: {size_limit}) have been reserved")

bsquizz (changes requested):

bonfire/openshift.py
```diff
@@ -327,3 +328,18 @@ def get_reservation(name=None, namespace=None, requester=None):
             return None
 
     return None
+
+
+def get_pool_size_limit(pool):
+    pool_data = get_json("namespacepool", name=pool)
+
+    return int(pool_data["spec"].get("sizeLimit", None)) if pool_data else None
```
bsquizz, line 336: I think you want to return `0` here instead of `None` so that the integer comparison back in bonfire.py doesn't fail:

```
    return int(pool_data["spec"].get("sizeLimit", 0)) if pool_data else 0
```

bsquizz (changes requested):

PR #256, opened 2022-10-20
feat: Add counter for image substitutions

SteveHNH:
Added a counter that we can add different substitions to. If the counter
comes back as `0` for image substitutions, then something went wrong and
we should error out

I added a counter as a dictionary in case we ever want to grab more detail out of it,
or use it for counting more stuff.

I can go back and make it a simple int if that's all we want.

I didn't see a good way to test this one in the unit tests but its been a while since
I've worked with python.

Signed-off-by: Stephen Adams <stephen.adams@redhat.com>

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -813,6 +813,7 @@ def _process(
     component_filter,
     local,
     frontends,
+    counter,
 ):
     apps_config = _get_apps_config(source, target_env, ref_env, local_config_path)
 
@@ -833,6 +834,7 @@ def _process(
         component_filter,
         local,
         frontends,
+        counter,
     )
     return processor.process()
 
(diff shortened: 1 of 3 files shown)


bsquizz (changes requested):
Good idea making the counter "future-proof" in case we want to count more things.

Generally I am thinking this:
1. Let's initialize the counter in TemplateProcessor, not in bonfire.py
2. We need to "initialize" the counter dictionary as well... something like:
```python
self.counter = {}
self.counter["image_tag_overrides"] = {}
for image in self.image_tag_overrides:
    self.counter["image_tag_overrides"][image] = 0
```
3. Let's throw the FatalError at the end of the 'process' function in TemplateProcessor

bsquizz (approved):
LGTM, tested it a bit as well.

PR #258, opened 2022-10-25
Run tests with python 3.6

bsquizz:
Since app-sre jenkins agents appear to run with py 3.6, we should run the PR tests with that version.


Victoremepunto:
@bsquizz  this won't work BC of the pinned version of setuptools

```  ERROR: No matching distribution found for setuptools==60.10.0```

bsquizz:
Will handle via other PRs

PR #259, opened 2022-10-27
!! Hotswap !!

gburges:
Adds a switch to CRCD if ephemeral api doesn't 200

### DIFF
--- a/cicd/bootstrap.sh
+++ b/cicd/bootstrap.sh
@@ -82,7 +82,7 @@ add_cicd_bin_to_path() {
 check_available_server() {
   echo "Checking connectivity to ephemeral cluster ..."
   (curl -s $OC_LOGIN_SERVER > /dev/null)
-  RET_CODE=$?
+  RET_CODE=1
   if [ $RET_CODE -ge 1 ]; then echo "Connectivity check failed"; fi
   return $RET_CODE
 }
--- a/cicd/bootstrap.sh
+++ b/cicd/bootstrap.sh
@@ -79,7 +79,27 @@ add_cicd_bin_to_path() {
   if ! command -v oc_wrapper; then export PATH=$PATH:${CICD_ROOT}/bin; fi
 }
 
+check_available_server() {
+  echo "Checking connectivity to ephemeral cluster ..."
+  (curl -s $OC_LOGIN_SERVER > /dev/null)
+  RET_CODE=$?
+  if [ $RET_CODE -ge 1 ]; then echo "Connectivity check failed"; fi
+  return $RET_CODE
+}
+
+# Hotswap based on availability
+login_to_available_server() {
+  if check_available_server; then
+    # log in to ephemeral cluster
+    oc_wrapper login --token=$OC_LOGIN_TOKEN --server=$OC_LOGIN_SERVER
+    echo "logging in to Ephemeral cluster"
+  else
+    # switch to crcd cluster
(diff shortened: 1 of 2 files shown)


cicd/bootstrap.sh
```diff
@@ -81,5 +81,13 @@ add_cicd_bin_to_path() {
 
 add_cicd_bin_to_path
 
-# log in to ephemeral cluster
-oc_wrapper login --token=$OC_LOGIN_TOKEN --server=$OC_LOGIN_SERVER
+# !!Hotswap based on availability
+status = $(curl -s -w 200 $OC_LOGIN_SERVER)
```
Victoremepunto, line 85: can we wrap this in a function to "get available server" and then be explicit about which server we're connecting to ? 

This way we can also handle better the event of "no cluster being available" too

gburges: added - be mindful, this is just for the pr_checks - mostly not for humans :)

cicd/bootstrap.sh
```diff
@@ -81,5 +81,19 @@ add_cicd_bin_to_path() {
 
 add_cicd_bin_to_path
 
-# log in to ephemeral cluster
-oc_wrapper login --token=$OC_LOGIN_TOKEN --server=$OC_LOGIN_SERVER
+# !!Hotswap based on availability
+login_to_available_server() {
+  status = $(curl -s -w 200 $OC_LOGIN_SERVER)
```
Victoremepunto, line 86: this will always display `200`. 

try 
```
curl -s -w 200 foobar
```

the -w ( `--write-out` ) will always output what you added to STDOUT

need to be updated to something else, perhaps just evaluating an url without output, like:  

```
openshift_cluster_available() {

    local SERVER_URL=$1

    curl -s "$SERVER_URL" >/dev/null
}

if openshift_cluster_available "$OC_LOGIN_SERVER"; then
# [ ... ] 
```

cicd/bootstrap.sh
```diff
@@ -81,5 +81,19 @@ add_cicd_bin_to_path() {
...
+  if status == 200; 
+    then
```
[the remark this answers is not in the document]
Victoremepunto: @gburges  if you're refactoring and using helper functions, feel free to move them top of the file please

Victoremepunto (changes requested):
mmm, on a second thought, this looks like needs more works

cicd/bootstrap.sh
```diff
@@ -79,7 +79,24 @@ add_cicd_bin_to_path() {
   if ! command -v oc_wrapper; then export PATH=$PATH:${CICD_ROOT}/bin; fi
 }
 
+check_available_server() {
+  curl -s $OC_LOGIN_SERVER > /dev/null
```
bsquizz, line 83: I think you're going to want to return the exit code of this command.


```suggestion
  echo "Checking connectivity to ephemeral cluster ..."
  (curl -s $OC_LOGIN_SERVER > /dev/null)
  RET_CODE=$?
  if [ $RET_CODE -ge 1 ]; then echo "Connectivity check failed"; fi
  return $RET_CODE
```

bsquizz (changes requested):

bsquizz (changes requested):

bsquizz (approved):

cicd/bootstrap.sh
Victoremepunto: the last command's return code executed within a function should be the function's return value, without an explicit return, AFAIK :thinking:

PR #260, opened 2022-10-27
Add Shellcheck action

Victoremepunto:


cicd/deploy_ephemeral_db.sh
```diff
@@ -10,37 +12,41 @@ DB_DEPLOYMENT_NAME="${DB_DEPLOYMENT_NAME:-$COMPONENT_NAME-db}"
...
+    "$COMPONENTS_ARG" \
+    "$COMPONENTS_RESOURCES_ARG" | oc_wrapper apply -f - -n "$NAMESPACE"
 
-bonfire namespace wait-on-resources $NAMESPACE --db-only
+bonfire namespace wait-on-resources "$NAMESPACE" --db-only
 
 # Set up port-forward for DB
 LOCAL_DB_PORT=$(python -c 'import socket; s=socket.socket(); s.bind(("", 0)); print(s.getsockname()[1]); s.close()')
-oc_wrapper port-forward svc/$DB_DEPLOYMENT_NAME $LOCAL_DB_PORT:5432 -n $NAMESPACE &
-PORT_FORWARD_PID=$!
```
Victoremepunto, line 33: is this okay to go @bsquizz  ? I don't see it used anywhere .

bsquizz: I am surprised but yeah, I see no reference to the stored pid anywhere so I think you are right -- it can go.

cicd/_common_deploy_logic.sh
```diff
@@ -112,17 +123,19 @@ function transform_arg {
     # transform components to "$1" options for bonfire
     options=""
     option="$1"; shift;
-    components="$@"
+    components="$*"
```
skarekrow, line 126: this can be a big change, along with the change below I'd be interested if any component behavior changes.

cicd/_common_deploy_logic.sh
```diff
@@ -112,17 +123,19 @@ function transform_arg {
...
+if [ -n "$COMPONENTS" ]; then
+    COMPONENTS_ARG=$(transform_arg --component $COMPONENTS)
+    export COMPONENTS_ARG
 fi
 
-if [ ! -z "$COMPONENTS_W_RESOURCES" ]; then
-    export COMPONENTS_RESOURCES_ARG=$(transform_arg --no-remove-resources $COMPONENTS_W_RESOURCES)
+if [ -n "$COMPONENTS_W_RESOURCES" ]; then
+    COMPONENTS_RESOURCES_ARG=$(transform_arg --no-remove-resources $COMPONENTS_W_RESOURCES)
+    export COMPONENTS_RESOURCES_ARG
```
skarekrow, line 140: I see this pattern all over, we should likely disable it. It's really quite ugly imo.

Victoremepunto: you have any suggestions on how you envision this should look like ? I understand why it is coded this way, essentially you want a list of args passed in to a command, so in order not to duplicate the `--no-remove-resources` bit that's the only way ... LMK Please

cicd/_common_deploy_logic.sh
Victoremepunto: This line was violating [SC2124](https://www.shellcheck.net/wiki/SC2124), "Assigning an array to a string" 

Instead of keeping the array, I thought it was simpler to remove the need of an array there and use a single string "$*" for the whole list of parameters , and then make use of the split by spaces without quoting the components list variable ( see lines 134  and 139 of the same script ) . 

@Ichimonji10  may have an opinion here too :)

cicd/_common_deploy_logic.sh
skarekrow: What I mean was it sets the variable and then exports it, whereas that happened on just one line previously. It's kind of like how our current stuff breaks our imports into single lines with our python projects.

cicd/_common_deploy_logic.sh
skarekrow: Sure, as long as it's accounted for. I was just highlighting that this is not normally a drop in replacement for `$@`.

cicd/examples/unit_test_example.sh
```diff
@@ -17,12 +17,13 @@ set -e
 # Evaluate the test result.
 
 # If you set up a python virtual env for your tests, move back to the bonfire virtual env...
+# shellcheck source=/dev/null
```
skarekrow, line 20: debug?

cicd/_common_deploy_logic.sh
```diff
@@ -28,45 +32,54 @@ trap_proxy teardown EXIT ERR SIGINT SIGTERM
...
+    local LOGS_DIR="${K8S_ARTIFACTS_DIR}/${NAMESPACE}/logs"
+
+    mkdir -p "$LOGS_DIR"
+
     # get array of pod_name:container1,container2,..,containerN for all containers in all pods
     echo "Collecting container logs..."
-    PODS_CONTAINERS=($(oc_wrapper get pods --ignore-not-found=true -n $ns -o "jsonpath={range .items[*]}{' '}{.metadata.name}{':'}{range .spec['containers', 'initContainers'][*]}{.name}{','}"))
-    for pc in ${PODS_CONTAINERS[@]}; do
+
+    IFS=" " read -r -a PODS_CONTAINERS <<< "$(oc get pods --ignore-not-found=true -n "$NAMESPACE" -o "jsonpath={range .items[*]}{' '}{.metadata.name}{':'}{range .spec['containers', 'initContainers'][*]}{.name}{','}")"
```
skarekrow, line 55: Normally you need to set IFS back, unless you really want it to be a space for the whole script.

cicd/_common_deploy_logic.sh
Victoremepunto: yeah well doing that in one line is actually violating https://www.shellcheck.net/wiki/SC2155

skarekrow: yeah I get their reasoning, it's a style thing for me, maybe it's not as ugly to others

cicd/_common_deploy_logic.sh
Ichimonji10: It sounds like you've got a handle on this already. Might any of the parameters ever contain whitespace? If so, this is problematic; if not, this is OK.

I don't really have time to read through this entire script and understand it, but this looks like it either normalizes or extracts parameters in `$@`. If so, an alternate approach is to use `getopt` (not `getopts`); see [getopt(1)](https://man.archlinux.org/man/core/util-linux/getopt.1.en), `/usr/share/doc/util-linux/`, or [this](https://github.com/pulp/pulp-fixtures/blob/f319a1831702e871605d82f260eb4560a6483f79/file/gen-fixtures.sh#L33-L61).

Victoremepunto: Yep - exactly what you said @Ichimonji10  - and nope, no spaces expected (they're supposed to be component identifiers without spaces on them so we should be good. 

About `get-opt` the thing is the parameters are read from a variable and not from a CLI option - getopt seems a bit over-engineered IMO, to basically process a list of space separated component ids, which is essentially all that's happening here. 

I appreciate your time with the review!

cicd/_common_deploy_logic.sh
Victoremepunto: yeah , the problem we're trying to avoid is to miss the expression evaluation result from the assignation action, which is always exit code 0 (that's what SC2155 is trying to help with), regardless of the syntax preventing us from getting away with a one-liner :)

cicd/_common_deploy_logic.sh
Victoremepunto: Oh my - thanks ! fixed

cicd/_common_deploy_logic.sh
```diff
@@ -28,45 +32,56 @@ trap_proxy teardown EXIT ERR SIGINT SIGTERM
...
+
+    local NAMESPACE="$1"
+    local LOGS_DIR="${K8S_ARTIFACTS_DIR}/${NAMESPACE}/logs"
+    local F_IFS="$IFS"
+
+    mkdir -p "$LOGS_DIR"
+
     # get array of pod_name:container1,container2,..,containerN for all containers in all pods
     echo "Collecting container logs..."
-    PODS_CONTAINERS=($(oc_wrapper get pods --ignore-not-found=true -n $ns -o "jsonpath={range .items[*]}{' '}{.metadata.name}{':'}{range .spec['containers', 'initContainers'][*]}{.name}{','}"))
```
bsquizz, line 47: IMO this was more easily understandable compared to what's now on line 56 w/ the IFS and 'read' stuff. What do you think?

cicd/build.sh
```diff
@@ -4,30 +4,30 @@
...
     fi
-
     # if this is a PR, set the tag to expire in 3 days
-    if [ ! -z "$ghprbPullId" ] || [ ! -z "$gitlabMergeRequestIid" ]; then
-        echo "LABEL quay.expires-after=${QUAY_EXPIRE_TIME}" >> $APP_ROOT/$DOCKERFILE
+    if [ -n "${ghprbPullId:-}" ] || [ -n "${gitlabMergeRequestIid:-}" ]; then
+        echo "LABEL quay.expires-after=${QUAY_EXPIRE_TIME}" >> "${APP_ROOT}/${DOCKERFILE}"
         # use IMAGE_TAG_LATEST later to detect extra tag to push
-        IMAGE_TAG_LATEST="$(cut -d "-" -f 1,2 <<< $IMAGE_TAG)-latest"
-        IMAGE_TAG_OPTS+=" -t ${IMAGE}:${IMAGE_TAG_LATEST}"
```
bsquizz, line 30: IMO the old line here using "+=" is more easily understandable compared to the new line 30

bsquizz (changes requested):
Just two lines that I think may be more confusing w/ the recommended changes compared to how they were written previously.

Do you know why shellcheck doesn't like those 2 lines? Is there some potentially "bad" about the way they were written?

Victoremepunto:
> Just two lines that I think may be more confusing w/ the recommended changes compared to how they were written previously.
> 
> Do you know why shellcheck doesn't like those 2 lines? Is there some potentially "bad" about the way they were written?

- About  the `_common_deploy_logic.sh` one, this is what shellcheck says 

```
In cicd/_common_deploy_logic.sh line 56:
    PODS_CONTAINERS=($(oc_wrapper get pods --ignore-not-found=true -n $ns -o "jsonpath={range .items[*]}{' '}{.metadata.name}{':'}{range .spec['containers', 'initContainers'][*]}{.name}{','}"))
                     ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).

```

I tried to apply what's suggested from [SC2207](https://www.shellcheck.net/wiki/SC2207), specifically: 
> If it outputs a line with multiple words (separated by spaces) ...
> \# For bash, uses temporary files
> IFS=" " read -r -a array <<< "$(mycommand)"

I've actually refactored a bit that very long line to try to improve readability, separating the containers per pod command into it's own function.

The Shellcheck comment is basically about reading a space separated input explicitly into an array.

- About the second one in `build.sh` I've properly used arrays and restored it.

LMK if this makes sense please. 

Thanks for the feedback ! (also @skarekrow as he gave feedback for point 1)

bsquizz (approved):

PR #263, opened 2022-11-10
Add support for --selenium to cji_smoke_test.sh

bsquizz:
Adds a couple more env vars to cji_smoke_test.sh that will allow us to run UI tests in ephemeral envs via PR checks


bsquizz:
Regression tested w/ rbac pr check

PR #264, opened 2022-11-10
Fix logs if sel container present and add args for deploying frontends

bsquizz:


bsquizz:
Regression tested w/ rbac

PR #270, opened 2022-11-16
Update to latest setuptools, refactor project config

Victoremepunto:
- Removes pin to specific setuptools version (**resources** dir seems to be fine, WRT comment in 005883122bed8c72b7e60cb4cc424b293eb9d05b)
- Removes `tests` from wheel to avoid them being distributed
- Refactor `setuptools_scm` and `setuptools` config sections to not to require setup.py
- Replace usage of `pkg_resources` with `importlib.metadata` for calculating `version`

### DIFF
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -13,7 +13,7 @@ from distutils.version import StrictVersion
 from pathlib import Path
 from urllib.parse import urlparse
 
-import pkg_resources
+from importlib.metadata import version, PackageNotFoundError
 import requests
 import yaml
 from cached_property import cached_property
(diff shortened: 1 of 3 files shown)


bonfire/utils.py
```diff
@@ -13,7 +13,7 @@
 from pathlib import Path
 from urllib.parse import urlparse
 
-import pkg_resources
+from importlib.metadata import version, PackageNotFoundError
```
maknop, line 16: It looks like from a quick google search that using `pkg_resources` is partially deprecated, was this the reason for the change or was there another reason?

Victoremepunto: Yep, exactly, I read about it while reading: https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources

Victoremepunto:
converting to draft, I think the tests should be included in sdist, not in wheel, need to check this first

Victoremepunto:
> converting to draft, I think the tests should be included in sdist, not in wheel, need to check this first

Ok confirmed, this should be good to go.

bsquizz (approved):

PR #273, opened 2022-11-18
Refactor pipelines

Victoremepunto:
- Refactor stages for building sdist and wheel first, then test on different Python versions
 - Heavily inspired by https://github.com/RonnyPfannschmidt/pytest-manual-marker - Thanks @RonnyPfannschmidt !
 - This should supersede #258


Victoremepunto:
@bsquizz  I'd like a word with you about `requirements` and how they're handled please in this project - I only used the `requirements.txt` file for the `build` job, and manually installed `pytest-mock` for the tests.

I think we should refactor this project to ... maybe only add the dependencies to `pyproject.toml` ? 

We also need to take a look at #274  - it is **not** caused by this PR though.

.github/workflows/tests.yml
```diff
@@ -6,46 +6,109 @@ on:
...
+        bonfire --help
+        bonfire config write-default
+    - name: pytest
+      run: pytest -sv
 
+  build_check:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+    - uses: actions/setup-python@v2
```
maknop, line 91: As someone who isn't as familiar with Github actions, I was wondering if it would be more beneficial to use `actions/setup-python@v4` instead? This may not be required for what we are needing it for, but I figured I'd ask. This way we are also keeping things updated. Looks like it is being used in the `test` job.

Victoremepunto: yep ! you're absolutely correct ! I saw an alert from the actions view that mentioned the Node version was going to make v2 to stop working

I've updated all the old references. Well, except the "flake8" action - I think I'm gonna refactor to not to use that at all - but on a follow-up PR

Thanks for mentioning it !

bsquizz:
> @bsquizz I'd like a word with you about `requirements` and how they're handled please in this project - I only used the `requirements.txt` file for the `build` job, and manually installed `pytest-mock` for the tests.

Yeah I don't think requirements.txt is used -- it was more of a reference for a set of known working pins. `setup.cfg` has the package dependencies set in `install_requires` and it is what ends up getting used in 'pip install' -- there is a set of dependencies defined for the 'test' extras so syntax like `pip install "crc-bonfire[test]"` or `pip install -e ".[test]"` could be used to install the test requirements like pytest-mock

bsquizz (approved):
This looks great. Thanks @Victoremepunto !

Issue #139, opened 2021-10-19
Add kubernetes context option


vkrizan:
Hi, 

Would it be possible to create a global option to use different kubernetes context?
It might be for example `--context`.

This could be handy if there are couple of other clusters that a user is connected to.
Workaround is to use `kubectl/oc config use-context ...` before `bonfire` execution.

Thanks!

PR #276, opened 2022-11-25
Optional namespace

vkrizan:
**Needs testing**

Use namespace from current context for majority of bonfire commands,
making the --namespace optional.

The deploy command tries to use current namespace, and if it is not
reserved or not owned it reserves a new one.  This is to keep backwards
compatiblity.

On each reservation a new namespace is set on current (or appropriate) context.

Adds --reserve option to deploy command to force a reservation of new
namespace.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -925,15 +925,7 @@ def _cmd_process(
 
 
 def _get_namespace(
-    requested_ns_name,
-    name,
-    requester,
-    duration,
-    pool,
-    timeout,
-    local,
-    force,
-    using_current=False
+    requested_ns_name, name, requester, duration, pool, timeout, local, force, using_current=False
 ):
 
     if not has_ns_operator():
@@ -950,8 +942,11 @@ def _get_namespace(
     reserved_new_ns = False
     if not ns:
         if using_current:
-            log.info("Current namespace '%s' could not be used (not reserved,"
-                     " expired, or not owned), reserving a new one", requested_ns_name)
+            log.info(
+                "Current namespace '%s' could not be used (not reserved,"
+                " expired, or not owned), reserving a new one",
+                requested_ns_name,
+            )
 
         ns = _check_and_reserve_namespace(name, requester, duration, pool, timeout, local, force)
         reserved_new_ns = True
@@ -963,9 +958,7 @@ def _check_and_use_namespace(requested_ns_name, using_current):
     if not has_ns_operator():
         _error(f"{NO_RESERVATION_SYS}")
 
-    log.debug(
-        "checking if namespace '%s' has been reserved via ns operator...", requested_ns_name
-    )
+    log.debug("checking if namespace '%s' has been reserved via ns operator...", requested_ns_name)
     operator_reservation = get_reservation(namespace=requested_ns_name)
     ns = None
     if operator_reservation:
@@ -1022,14 +1015,18 @@ def _check_and_reserve_namespace(name, requester, duration, pool, timeout, local
 @click.option(
     "--namespace",
     "-n",
-    help="Namespace (defaults to namespace from current context; "
-         "if not set, not reserved, or not owned, then bonfire will reserve a new one)",
+    help=(
+        "Namespace (defaults to namespace from current context; "
+        "if not set, not reserved, or not owned, then bonfire will reserve a new one)"
+    ),
     default=None,
 )
 @click.option(
     "--reserve",
-    help="Do not use current context's namespace and force reserve a new one. (keeps"
-         " the reservation on failure)",
+    help=(
+        "Do not use current context's namespace and force reserve a new one. (keeps"
+        " the reservation on failure)"
+    ),
     is_flag=True,
 )
 @click.option(
@@ -1095,8 +1092,15 @@ def _cmd_config_deploy(
         namespace = get_current_namespace()
 
     ns, reserved_new_ns = _get_namespace(
-        namespace, name, requester, duration, pool, timeout, local, force,
-        using_current=using_current
+        namespace,
+        name,
+        requester,
+        duration,
+        pool,
+        timeout,
+        local,
+        force,
+        using_current=using_current,
     )
 
     if import_secrets:
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -716,10 +716,10 @@ def _cmd_namespace_release(namespace, force, local):
         namespace = current_namespace_or_error()
 
     if not force:
-        _warn_before_delete()
         ns = Namespace(name=namespace)
         if not ns.owned_by_me:
             _warn_if_not_owned_by_me()
+        _warn_before_delete()
 
     release_reservation(namespace=namespace, local=local)
 
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1027,6 +1027,12 @@ def _check_and_reserve_namespace(name, requester, duration, pool, timeout, local
     default=None,
 )
 @click.option(
+    "--reserve",
+    help="Do not use current context's namespace and force reserve a new one. (keeps"
+         " the reservation on failure)",
+    is_flag=True,
+)
+@click.option(
     "--import-secrets",
     is_flag=True,
     help="Import secrets from local directory at deploy time",
@@ -1063,6 +1069,7 @@ def _cmd_config_deploy(
     no_remove_dependencies,
     single_replicas,
     namespace,
+    reserve,
     name,
     requester,
     duration,
@@ -1081,7 +1088,9 @@ def _cmd_config_deploy(
         _error("cluster does not have clowder operator installed")
 
     using_current = False
-    if not namespace and not conf.BONFIRE_BOT:
+    if reserve:
+        namespace = None
+    elif not namespace and not conf.BONFIRE_BOT:
         using_current = True
         namespace = get_current_namespace()
 
@@ -1107,7 +1116,7 @@ def _cmd_config_deploy(
 
     def _err_handler(err):
         try:
-            if not no_release_on_fail and reserved_new_ns:
+            if not no_release_on_fail and reserved_new_ns and not reserve:
                 # if we auto-reserved this ns, auto-release it on failure unless
                 # --no-release-on-fail was requested
                 log.info("releasing namespace '%s'", ns)
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -682,8 +682,8 @@ def _list_namespaces(available, mine, output):
 @click_exception_wrapper("namespace reserve")
 def _cmd_namespace_reserve(name, requester, duration, pool, timeout, local, force):
     """Reserve an ephemeral namespace"""
-    ns_name, _ = _get_namespace(None, name, requester, duration, pool, timeout, local, force)
-    click.echo(ns_name)
+    ns = _check_and_reserve_namespace(name, requester, duration, pool, timeout, local, force)
+    click.echo(ns.name)
 
 
 @namespace.command("release")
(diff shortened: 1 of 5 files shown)


bsquizz (changes requested):
I merged the ocviapy PR and released it -- can you update setup.cfg to require `ocviapy>=1.2.0` ?

https://github.com/RedHatInsights/bonfire/blob/master/setup.cfg#L33

bsquizz:
When this is merged, will this take care of #139 ?

vkrizan:
@bsquizz I've added a new requirement to get the fix I've just posted: https://github.com/RedHatInsights/ocviapy/pull/6

Additionally, I've added a fix for the `namespace describe` command in case the namespace is being released and cleaned up, and added a condition for deploy with `--reserve` to NOT drop the reservation on error.

bsquizz:
Ok, I will test this out. The py3.6 unit tests are failing for unrelated reasons.

bonfire/bonfire.py
```diff
@@ -702,6 +702,9 @@ def _cmd_namespace_release(namespace, force, local):
     if not has_ns_operator():
         _error(NO_RESERVATION_SYS)
 
+    if not namespace:
+        namespace = get_current_namespace()
```
bsquizz, line 706: I think we need to handle the case where `get_current_namespace` returns `None` here and print a fatal error message via the `_error()` func?

bonfire/bonfire.py
```diff
@@ -728,11 +731,14 @@ def _cmd_namespace_extend(namespace, duration, local):
...
```
bsquizz, line 735: Same here right?

bonfire/bonfire.py
```diff
@@ -903,65 +914,115 @@ def _cmd_process(
...
+    return reserve_namespace(name, requester, duration, pool, timeout, local)
 
 
 @main.command("deploy")
 @options(_process_options)
 @click.option(
     "--namespace",
     "-n",
-    help="Namespace to deploy to (if none given, bonfire will try to reserve one)",
+    help="Namespace (defaults to namespace from current conext; "
```
bsquizz, line 1016:
```suggestion
    help="Namespace (defaults to namespace from current context; "
```

bonfire/bonfire.py
```diff
@@ -903,65 +914,115 @@ def _cmd_process(
...
+         "if not set, not reserved or not owned, then bonfire will reserve a new one)",
```
bsquizz, line 1017:
```suggestion
         "if not set, not reserved, or not owned, then bonfire will reserve a new one)",
```

bonfire/bonfire.py
```diff
@@ -903,65 +914,115 @@ def _cmd_process(
...
     default=None,
 )
+@click.option(
+    "--reserve",
+    help="Do not use current context's namespace and force reserve a new one. (Keeps"
```
bsquizz, line 1022:
```suggestion
    help="Do not use current context's namespace and force reserve a new one. (keeps"
```

bonfire/namespaces.py
```diff
@@ -353,8 +356,11 @@ def describe_namespace(project_name: str):
         raise FatalError(f"namespace '{project_name}' was not reserved with namespace operator")
 
     routes = get_json("route", namespace=project_name)
-    host = get_fe_hostname(routes)
+    if not routes.get('items'):
+        raise FatalError(f"namespace '{project_name}' does not have Routes (is is being"
```
bsquizz, line 360: I don't think this is the best way to see if a namespace is in the process of getting cleaned up. You could possibly deploy templates that do not result in any Routes being created. Also, Routes are an OpenShift-specific thing and bonfire is technically supposed to work for deployments to both minikube or OpenShift -- on minikube there will be no Routes :)

A more reliable way might be if you see that the namespace is in 'Terminating' state when you look at its kubernetes object 'status' or something like that.

Is this kind of race condition a problem you run into frequently? If so maybe we can address this in a follow-up PR.

bonfire/bonfire.py
```diff
@@ -742,6 +748,8 @@ def _cmd_namespace_extend(namespace, duration, local):
 @options(_timeout_option)
 def _cmd_namespace_wait_on_resources(namespace, timeout, db_only):
     """Wait for rolled out resources to be ready in namespace"""
+    if not namespace:
+        namespace = get_current_namespace()
```
bsquizz, line 752: Same here?

bonfire/bonfire.py
```diff
@@ -750,9 +758,12 @@ def _cmd_namespace_wait_on_resources(namespace, timeout, db_only):
 
 
 @namespace.command("describe")
-@click.argument("namespace", required=True, type=str)
+@click.argument("namespace", required=False, type=str)
 def _describe_namespace(namespace):
     """Get current namespace info"""
+    if not namespace:
+        namespace = get_current_namespace()
```
bsquizz, line 765: Same here?

bonfire/bonfire.py
```diff
@@ -903,65 +914,115 @@ def _cmd_process(
...
             else:
-                log.debug("found existing ns operator reservation")
+                _error(msg)
 
-                if operator_reservation.get("status", {}).get("state") == "expired":
-                    _error(f"Reservation has expired for namespace '{requested_ns_name}'")
+        ns = Namespace(name=requested_ns_name)
+        if not ns.owned_by_me:
+            if using_current:
+                log.info("Current namespace '%s' is reserved by someone else, skipping",
```
bsquizz, line 975: "skipping" confused me a bit. I see how it works now -- looks like we'll get the log msg from line 943 printed next. I think let's just remove the "skipping" part from this log message.

```suggestion
                log.info("current namespace '%s' is reserved by someone else",
```

bonfire/bonfire.py
```diff
@@ -1016,8 +1078,16 @@ def _cmd_config_deploy(
     if not has_clowder():
         _error("cluster does not have clowder operator installed")
 
+    using_current = False
+    if reserve:
```
bsquizz, line 1082:
```suggestion
    if conf.BONFIRE_BOT or reserve:
```

I think we should also force bonfire to never run with 'using_current' mode if `conf.BONFIRE_BOT` is True -- that way our PR check behavior doesn't change.

Our CICD scripts run in 'bot' mode: https://github.com/RedHatInsights/bonfire/blob/master/cicd/bootstrap.sh#L21

This option is already used here for another purpose: https://github.com/RedHatInsights/bonfire/blob/d2dec39cca3e304efa69d72a9986344c0f2f45d7/bonfire/bonfire.py#L139

bonfire/bonfire.py
```diff
@@ -903,65 +914,115 @@ def _cmd_process(
...
+    local,
+    force,
+    using_current=False
+):
     reserved_new_ns = False
 
     if not has_ns_operator():
         if requested_ns_name:
             ns = Namespace(name=requested_ns_name)
+            return ns.name, requested_ns_name
```
bsquizz, line 933:
```suggestion
            return ns.name, False
```

The second value of the tuple should be 'True' or 'False' -- it indicates whether a namespace was newly reserved by bonfire when `_get_namespace` was invoked. In this area of the code, if there's no namespace operator, it is always 'False'

bonfire/bonfire.py
```diff
@@ -903,65 +914,115 @@ def _cmd_process(
...
-            if not operator_reservation:
-                _error(f"Namespace '{requested_ns_name}' has not been reserved yet")
+    ns = None
+    if requested_ns_name:
+        ns = _check_and_use_namespace(requested_ns_name, using_current)
+
+    if not ns:
+        if using_current:
+            log.info("Current namespace '%s' could not be used (not reserved,"
+                     " expired or not owned), reserving a new one", requested_ns_name)
```
bsquizz, line 944:
```suggestion
                     " expired, or not owned), reserving a new one", requested_ns_name)
```

bonfire/namespaces.py
```diff
@@ -283,6 +283,9 @@ def reserve_namespace(name, requester, duration, pool, timeout, local=True):
         pool,
     )
 
+    # set reserved namespace as current
+    set_current_namespace(ns_name)
```
bsquizz, line 287: Let's also disable this behavior when in 'bot' mode:

```suggestion
    if not conf.BONFIRE_BOT:
        set_current_namespace(ns_name)
```

bsquizz (changes requested):

bonfire/bonfire.py
bsquizz: (I am referring to the possibility that `get_current_namespace` could return `None`)

bonfire/bonfire.py
vkrizan: Good call! Added.

bonfire/bonfire.py
vkrizan: Yep, added.

bonfire/bonfire.py
vkrizan: Yep, added.

bonfire/bonfire.py
vkrizan: Added.

bonfire/bonfire.py
vkrizan: I've added it to the else if block, where it make sense. That would disable using namespace from current context if non is given for bot mode.

bonfire/namespaces.py
vkrizan: Makes sense. It would be better to not use any specialties like `BONFIRE_BOT` but rather have it universal and have an option for almost each action that needs to be configurable.

bonfire/namespaces.py
vkrizan: This fixed a `KeyError` from `get_fe_hostname`. I can keep this out of this PR.

The reproducer is:
# bonfire namespace reserve
# bonfire namespace release
# bonfire namespace describe

vkrizan: I've took it out of this PR. We can submit that separately to not block this one.

bonfire/bonfire.py
vkrizan: It is false from ` reserved_new_ns = False` (line 928). Do you want this to be explicit here?

vkrizan: Disregard, there is a wrong var used. Fixed.

bonfire/namespaces.py
bsquizz: Opened https://github.com/RedHatInsights/bonfire/issues/281 to track

bonfire/bonfire.py
bsquizz: Taking a closer look at this ... it should be OK if you type `bonfire deploy` without `--namespace` in bot mode... in the "current mode of operation" that would go ahead and reserve an ns for you and I think we want to maintain that behavior for any pr check scripts out there that call `bonfire deploy` without calling `bonfire namespace reserve` first.

bsquizz: Ok, double check me on this ... would this logic be right?:

If 'reserve' is true then we force reserve a namespace.

Otherwise, if there is no namespace explicitly set, we will try to use current namespace as long as we are not in 'bot' mode. I also tweaked this to call `current_namespace_or_error`. If we are in 'bot' mode, then `using_current` stays set to False and a new namespace is reserved just like how things have always worked.

```python
    using_current = False
    if reserve:
        namespace = None
    elif not namespace and not conf.BONFIRE_BOT:
        using_current = True
        namespace = current_namespace_or_error()

    ns, reserved_new_ns = _get_namespace(
        namespace, name, requester, duration, pool, timeout, local, force,
        using_current=using_current
    )
```

vkrizan: To avoid use of current namespace for bot let's change the condition but leave the rest as is:

```
    using_current = False
    if reserve:
        namespace = None
    elif not namespace and not conf.BONFIRE_BOT:
        using_current = True
        namespace = get_current_namespace()
```

`namespace` is `None` if not defined within bot command, so no force is required, and then it should follow original procedure with new namespace reservation.

Erroring out on no current namespace would break the original procedure with reservation, so keeping the original with `get_current_namespace()`.

bsquizz:
Testing this out and pushing a few more tweaks as I come across some log messages that could use improvement.

bsquizz:
Pushing a small log change for this scenario:

Namespace was released, but the context still was pointed to it. This caused the ns to be returned as 'None' and the log message printed that the current ns was 'None':

```
❯ bonfire deploy advisor
2022-12-15 16:21:22 [    INFO] [          MainThread] running (pid 141910): oc project -q 
2022-12-15 16:21:22 [    INFO] [          pid-141910]  |stderr| error: you do not have rights to view project "ephemeral-ery5nc" specified in your config or the project doesn't exist
2022-12-15 16:21:22 [ WARNING] [          MainThread] Non-zero return code ignored
2022-12-15 16:21:22 [    INFO] [          MainThread] Current namespace 'None' could not be used (not reserved, expired, or not owned), reserving a new one
```

bsquizz:
Found a bug where it seemed like the NS release was going to be allowed to proceed even though I did not own the NS:
```
❯ bonfire namespace release ephemeral-zkvglv

Deleting this reservation will also delete the associated namespace
Continue anyway? [y/N]:
```

Turns out we just needed to put the "warn if not owned by me" check up higher. (this bug may have been present even before this PR ...)

bsquizz (approved):
Will pre-release as v4.12.0b0 and test it in some PR checks

vkrizan:
Thank you @bsquizz

PR #286, opened 2023-02-03
Refactor build.sh to avoid modifying app's Dockerfile and instead set label when building the image

Victoremepunto:


Victoremepunto:
@bsquizz @gburges PTAL - tested in RedHatInsights/insights-puptoo#256

gburges (approved):

cicd/build.sh
```diff
@@ -16,21 +16,37 @@ set -e
...
+is_rhel7_host() {
+
+    local RELEASE_FILE='/etc/redhat-release'
+
+    [ -f "$RELEASE_FILE" ] && grep -q -i "release 7" "$RELEASE_FILE"
+}
+
 function build {
+
+    local IMAGE_TAG_LATEST=''
```
mshriver, line 36: This starts to break from the pattern of ENV variables being 'global' to the bonfire cicd scripts, where everything is exported in bootstrap.

`IMAGE_TAG_LATEST` is indeed currently only used in this function, so its currently a safe move to local, but I think it would be better to maintain the consistency of exporting that variable. Going as far as defining it in `bootstrap.sh`.

mshriver (approved):

cicd/build.sh
mshriver: Approving with this optional input, as the variable currently isn't exporting so I'm asking you to expand the scope of your intent.

bsquizz: I don't think IMAGE_TAG_LATEST was meant to be used outside of this func so I'm ok with it staying local.

bsquizz:
When reading the title of this PR I am expecting to see something change w/ respect to how the expiry label is being set. My understanding of the changes are that they are refactoring build.sh a bit (which is fine) -- but I don't see where anything is getting set as a build parameter -- are there updates to another file missing here? :)

cicd/build.sh
Victoremepunto: I hear you Mike, but as Brandon confirmed, I examined the code and IMAGE_TAG_LATEST is not used anywhere, nor meant to be used to configure tags for the image at top level, so that's why I refactored it to stay as local and prevent it from going to the external scope needlessly, hope you all agree.

cicd/build.sh
```diff
@@ -16,21 +16,37 @@ set -e
...
+
     if [ ! -f "$APP_ROOT/$DOCKERFILE" ]; then
         echo "ERROR: No $DOCKERFILE found"
         exit 1
     fi
 
-    # if this is a PR, set the tag to expire in 3 days and set arg 'TEST_IMAGE=true'
-    if [ ! -z "$ghprbPullId" ] || [ ! -z "$gitlabMergeRequestIid" ]; then
-        echo "LABEL quay.expires-after=${QUAY_EXPIRE_TIME}" >> $APP_ROOT/$DOCKERFILE
```
Victoremepunto, line 27: @bsquizz  the reason for the PR title being "set as build parameter" is because I removed this particular line(old line 27), that edited the app's Dockerfile, and instead used is as an option to the `docker` / `podman` *build* command, just a few lines after it in (the new line 44), where I add it to the `CMD_OPTS` variable. So the expected build command should look like:

```
podman/docker build 
 -t ${IMAGE}:${IMAGE_TAG} 
 -t ${IMAGE}:${IMAGE_TAG_LATEST}
 --label quay.expires-after=${QUAY_EXPIRE_TIME}
 --build-arg TEST_IMAGE=true -t 
```

not to be confused with a *build-arg* . I have renamed the PR to properly reflect this change, hope it's clearer then . In other words, my intention here is to avoid to touch the app's Dockerfile and instead set the label as a parameter to `build`. Hope that makes sense.

bsquizz: I see! Clearly I missed that :D

bsquizz (approved):

PR #287, opened 2023-02-06
Invoke pip with `python3 -m pip`

catastrophe-brandon:
When invoking pip directly, I was seeing errors in the Jenkins job. When I switched to invoking with `python3 -m pip` I was able to get beyond my errors. It sounds like this is the recommended way to invoke pip now, so I'd advocate for adopting this in bootstrap.sh unless there are reasonable objections.


infinitewarp (approved):
LGTM

bsquizz (approved):

PR #298, opened 2023-03-22
Dynamically get namespace pools

### LINKED ISSUE #261: Retrieve different pool types from environment
Currently, the available `pool` [types](https://github.com/RedHatInsights/bonfire/blob/de03c7e245aa853642743ccfc9fa3e4a1cfa5df1/bonfire/config.py#L19) are hardcoded, it would be interesting to be able to fetch them from the environment of the target cluster, if possible.

gburges:
Special credits for @mjholder who rubber ducked/helped me with telling me print doesn't return values

Aims to fix https://github.com/RedHatInsights/bonfire/issues/261

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -24,6 +24,7 @@ from bonfire.namespaces import (
 from bonfire.openshift import (
     check_for_existing_reservation,
     find_clowd_env_for_ns,
+    get_namespace_pools,
     get_reservation,
     has_clowder,
     has_ns_operator,
@@ -232,8 +233,8 @@ _ns_reserve_options = [
     ),
     click.option(
         "--pool",
-        type=click.Choice(conf.NAMESPACE_POOLS, case_sensitive=False),
-        default="default",
+        type=click.Choice(str),
+        default=conf.DEFAULT_NAMESPACE_POOL,
         show_default=True,
         help="Specifies the pool type name",
     ),
@@ -865,7 +866,7 @@ def _process(
 @pool.command("list")
 def _cmd_pool_types():
     """List all pool types"""
-    click.echo("\n".join(conf.NAMESPACE_POOLS))
+    click.echo(get_namespace_pools)
 
 
 @main.command("process")
@@ -999,6 +1000,10 @@ def _check_and_use_namespace(requested_ns_name, using_current):
 def _check_and_reserve_namespace(name, requester, duration, pool, timeout, local, force):
     if not has_ns_operator():
         _error(f"{NO_RESERVATION_SYS}")
+
+    get_namespace_pools()
+    if pool not in get_namespace_pools():
+        _error(f"namespace pool '{pool}' does not exist on this cluster")
 
     log.debug("checking if requester already has another namespace reserved...")
     requester = requester if requester else _get_requester()
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -233,7 +233,7 @@ _ns_reserve_options = [
     ),
     click.option(
         "--pool",
-        type=click.Choice(str),
+        type=str,
         default=conf.DEFAULT_NAMESPACE_POOL,
         show_default=True,
         help="Specifies the pool type name",
@@ -866,7 +866,7 @@ def _process(
 @pool.command("list")
 def _cmd_pool_types():
     """List all pool types"""
-    click.echo(get_namespace_pools)
+    click.echo("\n".join(get_namespace_pools()))
 
 
 @main.command("process")
@@ -929,7 +929,6 @@ def _cmd_process(
 def _get_namespace(
     requested_ns_name, name, requester, duration, pool, timeout, local, force, using_current=False
 ):
-
     if not has_ns_operator():
         if requested_ns_name:
             ns = Namespace(name=requested_ns_name)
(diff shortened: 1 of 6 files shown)


bonfire/config.py
```diff
@@ -16,7 +16,7 @@
 DEFAULT_ENV_PATH = get_config_path().joinpath("env")
 DEFAULT_SECRETS_DIR = get_config_path().joinpath("secrets")
 
-NAMESPACE_POOLS = ["default", "minimal", "managed-kafka", "real-managed-kafka"]
+with os.popen("oc get namespacepool --no-headers| cut -d' ' -f1") as f: NAMESPACE_POOL = f.readlines()
```
bsquizz, line 19: A couple things to note here:
1. You should be able to call `from ocviapy import get_json` and then call `get_json("namespacepool")` function instead, which is essentially going to run `oc get namespacepool -o json` and return the result to you in a dictionary which you can then analyze.
2. I would not load the namespace pool on the top-level line of config here because that will cause this code to run any time bonfire runs. For example, even if you ran `bonfire --help` this could would get invoked.

Here is what I would suggest:
1. Get rid of `NAMESPACE_POOLS` from config and instead have just a `DEFAULT_NAMESPACE_POOL = "default"`

2. Add a `get_ns_pools` function to https://github.com/RedHatInsights/bonfire/blob/master/bonfire/openshift.py -- you can cache the result of this function (since the namespace pools are unlikely to change in the middle of a bonfire run) so use this same decorator: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/openshift.py#L23 -- if there are no namespace pools found this function should probably return `[]`

3. I would then call the new function you wrote at this point: https://github.com/RedHatInsights/bonfire/blob/69b3f34b6f303fa182e7765fd329aa18d47785a1/bonfire/bonfire.py#L1002

to ensure that 'pool' is in the valid list of pools found on the cluster

4. The namespace pool list command will also need to be changed to call your new function instead of just printing `config.NAMESPACE_POOLS`:

https://github.com/RedHatInsights/bonfire/blob/69b3f34b6f303fa182e7765fd329aa18d47785a1/bonfire/bonfire.py#L868

5. And you will need to edit the 'type' here back to simply `type=str`

6. You'll have to update this unit test: https://github.com/RedHatInsights/bonfire/blob/69b3f34b6f303fa182e7765fd329aa18d47785a1/tests/test_bonfire.py#L271

Mock out the return value of your new 'get_ns_pools' function similar to how 'has_ns_operator' has been mocked out here: https://github.com/RedHatInsights/bonfire/blob/69b3f34b6f303fa182e7765fd329aa18d47785a1/tests/test_bonfire.py#L255

bsquizz (changes requested):

bonfire/openshift.py
```diff
@@ -44,6 +44,14 @@ def get_console_url():
         return url
 
 
+@functools.lru_cache(maxsize=None, typed=False)
+def get_namespace_pools():
+    namespace_pools = get_json("namespacepool")
```
bsquizz, line 49: You probably want this function to simply return a list of names.

i.e.:

```python
    namespace_pools = get_json("namespacepool")
    return [pool["metadata"]["name"] for pool in namespace_pools]
```

tests/test_bonfire.py
```diff
@@ -254,6 +254,7 @@ def test_ns_reserve_flag_timeout(mocker, caplog, user: str, namespace: str, time
     mocker.patch("bonfire.bonfire.get_pool_size_limit", return_value=0)
     mocker.patch("bonfire.bonfire.has_ns_operator", return_value=True)
     mocker.patch("bonfire.namespaces.whoami", return_value=user)
+    mocker.patch("bonfire.bonfire.get_namespace_pools", return_value=conf.DEFAULT_NAMESPACE_POOL)
```
bsquizz, line 257: You need to mock this to return a list of strings. E.x. `return_value=["my", "fake", "pools"]`

tests/test_bonfire.py
```diff
@@ -277,7 +278,7 @@ def test_pool_list_command(caplog):
 
     output = [r for r in result.output.split("\n") if r != ""]
 
-    assert output == conf.NAMESPACE_POOLS
+    assert len(output) > 0
```
bsquizz, line 281: Assert that the output matches. Ex.:

```python
    expected = "\n".join(["my", "fake", "pools"])
    assert output == expected
```

bonfire/bonfire.py
```diff
@@ -1000,6 +1001,8 @@ def _check_and_reserve_namespace(name, requester, duration, pool, timeout, local
     if not has_ns_operator():
         _error(f"{NO_RESERVATION_SYS}")
 
+    get_namespace_pools()
```
bsquizz, line 1004: Here you should check the result of this function call to make sure the 'pool' that the user is trying to use is valid.

For ex:

```python
    if pool not in get_namespace_pools():
        _error(f"namespace pool '{pool}' does not exist on this cluster")
```

bonfire/bonfire.py
```diff
@@ -865,7 +866,7 @@ def _process(
 @pool.command("list")
 def _cmd_pool_types():
     """List all pool types"""
-    click.echo("\n".join(conf.NAMESPACE_POOLS))
+    click.echo(get_namespace_pools)
```
bsquizz, line 869:
```suggestion
    if not has_ns_operator():
        _error(f"{NO_RESERVATION_SYS}")
    click.echo("\n".join(get_namespace_pools()))
```

bsquizz (changes requested):

bsquizz (approved):

PR #301, opened 2023-04-03
Merge local app components with app-sre ones

### LINKED ISSUE #300: Support overriding a component via a yaml
Currently components are not patched, but are instead used in their entirety.

This ticket captures some of that behavior: https://issues.redhat.com/browse/RHCLOUD-25035

We'll need to update the readme and the area around here: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/local.py#L58

skarekrow:
Fixes #300

This should obviate the original behavior of merging the local apps in totality and instead patch the apps specifed with the component overrides.

Special thanks to @rsnyman and @bsquizza for their help with this.


bonfire/bonfire.py
```diff
@@ -793,9 +795,21 @@ def _get_apps_config(source, target_env, ref_env, local_config_path):
...
+                continue
+            for idx, component in enumerate(app_cfg.get("components", [])):
+                matched_components = list(
+                    filter(
+                        lambda c: component["name"] == c["name"], local_apps[app_name]["components"]
+                    )
+                )
 
+                # There can be only one
+                if len(matched_components) == 1:
```
bsquizz, line 811: What if you want to add a new component in your local apps config that is not yet in app-interface? The match result would be `0` and I think in that case we could just append the local `component` to `app_cfg["components"]` right?

Should we raise an error if `len(matched_components) > 1` ?

bsquizz (changes requested):

bonfire/bonfire.py
skarekrow: Well if it's not in app-interface, this code won't run right? That's covered by the if statement below? I felt like softly not doing anything instead of dying was the right course, but up to you if we want to raise.

bsquizz: I am thinking of this scenario...

Local config looks like this:
```
apps:
  - name: ingress
    components:
    - name: some-new-component
```

and an app named `ingress` exists in app-interface and it has other components defined already.

What if 'some-new-component' does not exist yet in app-interface? Right now I think your logic would filter this component out. You probably still want 'some-new-component' to go in if you type `bonfire deploy ingress` right?

bsquizz: Thinking on this further ... what if you have an entire 'app' in your local apps that is not yet in app-interface? The old logic would add it to your config. This new logic would omit it.

bsquizz: I pushed some new logic that should address these cases

bsquizz:
While I was at it I also renamed the 'local' source mode to 'file' because 'local' is a misleading name.

The local config is always parsed, even in 'appsre' mode. Really, what 'local' mode meant was that bonfire would fetch its apps config using the file defined at the `appsFile` in config.yaml instead of using app-sre. That mode is not currently used in the platform tooling so I do not think renaming it will have large impact, and that mode may prove useful again in the future so I kept the ability to define an `appsFile` in the code.

Whether your source of truth is app-sre or the 'apps file', the `--local-config-method` will operate the same way when it comes to merging your locally defined `apps` with the remotely defined ones.

bsquizz:
When we release this, we should tag this as `v5` since this is technically making a "breaking change" -- although I suspect it will impact very few people... and for the people it does impact, they just need a heads up about the new CLI flag if they desire to return to the old config merging behavior.

Victoremepunto:
@skarekrow @bsquizz  Can we add some E2E tests to keep this behavior away of future possible regressions ? they'll also help understand the behavior, I believe looking at a simplified use case with expectations is clearer than anything. 

Thanks!

bsquizz:
Latest commit keeps this all backward-compatible:
1. default merge method remains 'override' as it was before
2. the source named 'local' is kept around for now and there's a warning indicating that it is deprecated and has been renamed to 'file'

bsquizz:
> Latest commit keeps this all backward-compatible:
> 
> 1. default merge method remains 'override' as it was before
> 2. the source named 'local' is kept around for now and there's a warning indicating that it is deprecated and has been renamed to 'file'

Commit `10b78b7` reverted since we'll get approval for the change in default behavior via ADR process.

Issue #265, opened 2022-11-11
Bonfire should detect earlier it cannot reach template source environment


Victoremepunto:
If there's no connectivity to reach the template's source URL - the bonfire process takes up to 5 minutes to fail over timeout. 

I guess a preemptive test of the template source URL could just figure this out and fail quicker. An excerpt of the logs :

```
# bonfire deploy rbac -n $EE
2022-11-11 09:50:42 [ WARNING] [          MainThread] split() requires a non-empty pattern match.
2022-11-11 09:50:50 [    INFO] [          MainThread] processing app templates...
2022-11-11 09:50:50 [    INFO] [          MainThread] reading config from: *********/config.yaml
2022-11-11 09:50:50 [    INFO] [          MainThread] fetching apps config using source: ****, target env: **********
2022-11-11 09:55:10 [   ERROR] [          MainThread] hit unexpected error!
Traceback (most recent call last):
 # Traceback follows ....
```

bsquizz:
Need more context here on what the traceback and error is. I thought we sorted out this connectivity check issue here:

https://github.com/RedHatInsights/bonfire/pull/257

I am wondering if in your case, you had a successful DNS resolution but the actual network connection attempt is what timed out?

Victoremepunto:
@bsquizz  I was NOT logged into the VPN , however it didn't stop from getting 5 minutes of wait until it failed.

tried with:

```bash

$ time bonfire process compliance -n foo

# Stacktrace goes here

real	4m20.531s
user	0m0.320s
sys	0m0.046s
```

bsquizz:
Do you have the full stack trace? I am interested to see what area of code we were in at this point

PR #314, opened 2023-07-25
fix(utils): Timeout for qontract URL connection check

Victoremepunto:
- closes RHCLOUD-23183 and #265

### DIFF
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -566,7 +566,7 @@ def _check_connection(hostname, port=443, timeout=5):
             test_connection_socket.connect((hostname, port))
 
     except TimeoutError as err:
-        err.args = (f"Unable to connect to '{hostname} on port {port} after {timeout} seconds'. "
+        err.args = (f"Unable to connect to '{hostname}' on port {port} after {timeout} seconds. "
                     f"Check network connection (is VPN needed?)",)
         raise
 
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -33,10 +33,6 @@ class FatalError(Exception):
     pass
 
 
-class ConnectionError(FatalError):
-    pass
-
-
 def get_config_path():
     xdg_config_home = os.environ.get("XDG_CONFIG_HOME")
     if xdg_config_home:
@@ -555,28 +551,29 @@ def hms_to_seconds(s):
 
 
 @lru_cache(maxsize=None)
-def check_hostname(hostname):
+def __check_connection(hostname, port=443, timeout=5):
     """
     Check connection makes sure a connection is available to a given hostname.
 
-    Function is cached so that we only check a hostname once.
+    Function is cached so that we onl
+    y check a hostname once.
     """
     log.debug("checking connection to '%s'", hostname)
 
     try:
-        socket.gethostbyname(hostname)
-    except socket.gaierror:
-        raise ConnectionError(
-            f"Unable to resolve hostname '{hostname}'. Check network connection (is VPN needed?)"
-        )
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as test_connection_socket:
+            test_connection_socket.settimeout(timeout)
+            test_connection_socket.connect((hostname, port))
+
+    except TimeoutError as err:
+        err.args = (f"Unable to connect to '{hostname} on port {port} after {timeout} seconds'. "
+                    f"Check network connection (is VPN needed?)",)
+        raise
 
 
 def check_url_connection(url):
-    try:
-        parsed_url = urlparse(url)
-    except ValueError as err:
-        raise ValueError(f"invalid url '{url}': {err}")
+    parsed_url = urlparse(url)
+    if not (hostname := parsed_url.netloc):
+        raise ValueError(f"Cannot extract hostname from URL: '{url}'")
 
-    hostname = parsed_url.netloc
-
-    return check_hostname(hostname)
+    return __check_connection(hostname)
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -555,8 +555,7 @@ def _check_connection(hostname, port=443, timeout=5):
     """
     Check connection makes sure a connection is available to a given hostname.
 
-    Function is cached so that we onl
-    y check a hostname once.
+    Function is cached so that we only check a hostname once.
     """
     log.debug("checking connection to '%s'", hostname)
 
@@ -575,5 +574,4 @@ def check_url_connection(url):
     parsed_url = urlparse(url)
     if not (hostname := parsed_url.netloc):
         raise ValueError(f"Cannot extract hostname from URL: '{url}'")
-
     _check_connection(hostname)
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -564,10 +564,9 @@ def _check_connection(hostname, port=443, timeout=5):
             test_connection_socket.settimeout(timeout)
             test_connection_socket.connect((hostname, port))
 
-    except TimeoutError as err:
-        err.args = (f"Unable to connect to '{hostname}' on port {port} after {timeout} seconds. "
-                    f"Check network connection (is VPN needed?)",)
-        raise
+    except TimeoutError:
+        raise FatalError(f"Unable to connect to '{hostname}' on port {port} after {timeout} "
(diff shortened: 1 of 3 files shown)


bonfire/utils.py
```diff
@@ -555,28 +551,29 @@ def hms_to_seconds(s):
...
 
 @lru_cache(maxsize=None)
-def check_hostname(hostname):
+def __check_connection(hostname, port=443, timeout=5):
     """
     Check connection makes sure a connection is available to a given hostname.
 
-    Function is cached so that we only check a hostname once.
+    Function is cached so that we onl
+    y check a hostname once.
```
bsquizz, line 559: Again being picky here but can we fix this mid-word newline too? :)

process-rbac-foo.sh
```diff
@@ -0,0 +1,2 @@
+
+bonfire process rbac -n foo
```
bsquizz, line 2: Was this file accidentally included?

bonfire/utils.py
```diff
@@ -555,28 +551,29 @@ def hms_to_seconds(s):
...
-    except socket.gaierror:
-        raise ConnectionError(
-            f"Unable to resolve hostname '{hostname}'. Check network connection (is VPN needed?)"
-        )
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as test_connection_socket:
+            test_connection_socket.settimeout(timeout)
+            test_connection_socket.connect((hostname, port))
 
+    except TimeoutError as err:
+        err.args = (f"Unable to connect to '{hostname} on port {port} after {timeout} seconds'. "
```
bsquizz, line 569: Minor tweak on the closing quote here:

```suggestion
        err.args = (f"Unable to connect to '{hostname}' on port {port} after {timeout} seconds. "
```

bonfire/utils.py
```diff
@@ -555,28 +551,29 @@ def hms_to_seconds(s):
...
-    Function is cached so that we only check a hostname once.
+    Function is cached so that we onl
+    y check a hostname once.
     """
     log.debug("checking connection to '%s'", hostname)
 
     try:
-        socket.gethostbyname(hostname)
-    except socket.gaierror:
-        raise ConnectionError(
```
bsquizz, line 569: The reason that the old code raised a `FatalError` here is that it prints a cleaner message (`"ERROR: <message>"`) and calls `sys.exit(1)` without printing the entire traceback which can make the logs less overwhelming. I feel that single error message is probably going to tell the whole story to the end user and they won't need to see the whole traceback -- so you could `raise FatalError` instead of re-raising TimeoutError on line 571. What do you think?

bonfire/utils.py
```diff
@@ -555,28 +551,29 @@ def hms_to_seconds(s):
...
-        raise ValueError(f"invalid url '{url}': {err}")
 
-    hostname = parsed_url.netloc
+def check_url_connection(url):
+    parsed_url = urlparse(url)
+    if not (hostname := parsed_url.netloc):
+        raise ValueError(f"Cannot extract hostname from URL: '{url}'")
 
-    return check_hostname(hostname)
+    return __check_connection(hostname)
```
bsquizz, line 579: This is picky, but technically nothing is done with the return value of `__check_connection` -- and that function doesn't return anything ... so I say why bother to `return` :)

```suggestion
    _check_connection(hostname)
```

bonfire/utils.py
```diff
@@ -555,28 +551,29 @@ def hms_to_seconds(s):
 
 
 @lru_cache(maxsize=None)
-def check_hostname(hostname):
+def __check_connection(hostname, port=443, timeout=5):
```
bsquizz, line 554: Being picky, but does this need the double `_`?

```suggestion
def _check_connection(hostname, port=443, timeout=5):
```

bsquizz (changes requested):
A few nitpicks. Nice catch on with the `netloc` parsing, and thanks for adding unit tests!

bonfire/utils.py
Victoremepunto: non picky, this is a mistake from the editor formatting, my bad. Thanks for spotting this.

process-rbac-foo.sh
Victoremepunto: it was, sorry - I use this script locally for some sort of E2E testing occasionally with a run config on the IDE

bonfire/utils.py
Victoremepunto: Yeah we can do that - I've updated it to catch the Timeout and raise FatalError instead.

bonfire/utils.py
Victoremepunto: Agreed!

bonfire/utils.py
Victoremepunto: Thanks!

bonfire/utils.py
Victoremepunto: Not picky, PEP8 specifies one leading underscore for non-public methods. Thanks for raising it

bsquizz:
I disconnected from VPN and tested this out with `bonfire process engine` and ran into this:
```
2023-07-26 16:38:08 [    INFO] [          MainThread] processing component engine
2023-07-26 16:38:08 [   ERROR] [          MainThread] failed to fetch template file for engine
ERROR: [Errno -2] Name or service not known
```

Take a look at my latest commit which should also catch a DNS lookup error and let me know if you're OK with it.

bsquizz:
Also, `socket.connect` can raise `OSError` under certain cases (https://docs.python.org/3/library/socket.html#exceptions) so I included `OSError` in addition to `TimeoutError`

tests/test_utils.py
```diff
@@ -68,3 +69,35 @@ def test_hms_to_seconds(seconds: str, expected: int):
...
+    socket_mock = socket_library_mock.return_value.__enter__.return_value
+    socket_mock.connect.side_effect = TimeoutError("timed out!")
+
+    with pytest.raises(FatalError, match=r"Unable to connect to.*after.*seconds.*is VPN needed.*"):
+        check_url_connection("https://timingout.com")
+
+
+def test_ip_timeout():
+    with pytest.raises(FatalError, match="Unable to connect to.*after 1 seconds.*is VPN needed.*"):
+        check_url_connection("https://10.255.255.1", timeout=1)
```
Victoremepunto, line 98: I'm not sure about this, the moment that IP responds back the test is going to fail - however I don't see how "easily" we could mock that. Besides it shouldn't be necessary to actually run a network request externally against the default gateway IP (I think this is what you were trying). Main reason is, if this ever changes, the exception captured in the test is going to fail with a ConnectionRefused - Which - will be shadowed by the OSError and still make the test to pass.

I must say I tried creating a "small" socket server in a fixture so we didn't mock but actually had a socket listening to force the timeout with a custom handler, but that didn't go well. I was never able to make it work (and even if I did, it was going to require either threading or using asyncio, which was building up to be a "way to complex" for the task at hand so it didn't make sense even if I pulled it off)

Anyways - I think I just wanted to make my point, we can leave it as it is - but let me know your thoughts please.

tests/test_utils.py
```diff
@@ -68,3 +69,35 @@ def test_hms_to_seconds(seconds: str, expected: int):
...
+
+
+def test_url_connection_dns_lookup_fails():
+    with pytest.raises(FatalError, match=r".*DNS lookup failed.*"):
+        check_url_connection("https://baddomain.test")
```
Victoremepunto, line 103: I don't think there's a reason not to mock this and use a side effect to raise a `socket.gaierror`, this way we don't need to get creative with domain names if `baddomain.test` ever exists :)

Victoremepunto (commented):
Couple of notes - LMK your take and let's get this merged

tests/test_utils.py
bsquizz: Looks like a domain with `.invalid` should never resolve as per the RFC's: https://serverfault.com/a/846523 -- I'll switch to that.

tests/test_utils.py
Victoremepunto: Ah - okay, found the [RFC](https://datatracker.ietf.org/doc/html/rfc1918#section-3) - thank you

tests/test_utils.py
Victoremepunto: Nice ! Thanks

PR #315, opened 2023-08-01
Remove deprecated appsfile-config

Victoremepunto:
- Closes RHCLOUD-18925
- Adds tests for local processing
- Comment out all sections from default config file template


bsquizz:
Commenting out the `appsFile` from default config has been handled by #334 -- also a lot of new unit tests related to the local config file went in with #301 -- if we have any gaps in coverage let's just open a new PR to add them to that file

PR #317, opened 2023-08-01
Update list of frontend dependencies

bsquizz:
Some of the Frontend configurations have been split out of `frontend-configs` (chrome, landing, and dashboard) into their own separate deployment config. This updates the list.


bsquizz:
Need to fix the `sed-frontend` config in app-interface before merging this

PR #320, opened 2023-08-22
Override frontend auto-added dependencies with BONFIRE_FRONTEND_DEPENDENCIES env var

bsquizz:
This will allow teams (like HAC team) who don't need all these components pulled in to be able to customize the list by setting the env var, e.g.:

```
BONFIRE_FRONTEND_DEPENDENCIES="chrome-service,frontend-configs" bonfire deploy <args> --frontends=true
```

You can also pass an empty value which will cause no frontend dependencies to be pulled in automatically, e.g:
```
BONFIRE_FRONTEND_DEPENDENCIES="" bonfire deploy <args> --frontends=true
```

### DIFF
--- a/bonfire/config.py
+++ b/bonfire/config.py
@@ -54,7 +54,7 @@ BONFIRE_NS_REQUESTER = os.getenv("BONFIRE_NS_REQUESTER")
 # set to true when bonfire is running via automation using a bot acct (not an end user)
 BONFIRE_BOT = os.getenv("BONFIRE_BOT")
 
-AUTO_ADDED_FRONTEND_DEPENDENCIES = (
+DEFAULT_FRONTEND_DEPENDENCIES = (
     "chrome-service",
     "landing-page-frontend",
     "insights-chrome",
@@ -63,8 +63,21 @@ AUTO_ADDED_FRONTEND_DEPENDENCIES = (
     "rbac-frontend",
     "host-inventory",
     "host-inventory-frontend",
-    "unleash-proxy"
+    "unleash-proxy",
 )
+
+
+def _parse_frontend_dependencies():
+    env_var = os.getenv("BONFIRE_FRONTEND_DEPENDENCIES")
+    if isinstance(env_var, str):
+        if env_var.strip() == "":
+            return set()
+        return set(env_var.split(","))
+
+    return set(DEFAULT_FRONTEND_DEPENDENCIES)
+
+
+AUTO_ADDED_FRONTEND_DEPENDENCIES = _parse_frontend_dependencies()
 
 
 def write_default_config(outpath=None):
--- a/bonfire/config.py
+++ b/bonfire/config.py
@@ -67,17 +67,20 @@ DEFAULT_FRONTEND_DEPENDENCIES = (
 )
 
 
-def _parse_frontend_dependencies():
+def _get_auto_added_frontend_dependencies():
     env_var = os.getenv("BONFIRE_FRONTEND_DEPENDENCIES")
-    if isinstance(env_var, str):
-        if env_var.strip() == "":
-            return set()
-        return set(env_var.split(","))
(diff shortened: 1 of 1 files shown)


bonfire/config.py
```diff
@@ -54,17 +54,30 @@
...
     "rbac",
     "rbac-frontend",
     "host-inventory",
     "host-inventory-frontend",
-    "unleash-proxy"
+    "unleash-proxy",
 )
 
 
+def _parse_frontend_dependencies():
```
Victoremepunto, line 68: suggest to rename as `_get_auto_added_frontend_dependencies`, or if you want to refactor where it is used in `process` too, maybe rename the variable as `FRONTEND_DEPENDENCIES` and the method as  `_get_frontend_dependencies`.

bonfire/config.py
```diff
@@ -54,17 +54,30 @@
...
+    env_var = os.getenv("BONFIRE_FRONTEND_DEPENDENCIES")
+    if isinstance(env_var, str):
+        if env_var.strip() == "":
+            return set()
+        return set(env_var.split(","))
+
+    return set(DEFAULT_FRONTEND_DEPENDENCIES)
```
Victoremepunto, line 75: I'm wondering, maybe it's an enhancement but we're changing the type of `AUTO_ADDED_FRONTEND_DEPENDENCIES` from `tuple` to `set` , as long as we're fine with it then OK.

Also, I think this needs to strip out the spaces in case someone uses spaces:

```
BONFIRE_FRONTEND_DEPENDENCIES='bar ,   baz   , foo' 
```
Also, since `os.getenv` always returns either a `None` if not defined, or a `str` if defined, one could rewrite as



```suggestion

def _parse_frontend_dependencies():
    env_var = os.getenv("BONFIRE_FRONTEND_DEPENDENCIES")
    result = set()
  
    if env_var is None:
        result = set(DEFAULT_FRONTEND_DEPENDENCIES)
    else:
        for app in (env_var.split(",")):
            result.add(app.strip())

    return result

    return set(DEFAULT_FRONTEND_DEPENDENCIES)
```

Victoremepunto (commented):
Spotted a couple of things may worth revisiting, LMK.

bonfire/config.py
bsquizz: Had to tweak slightly but adopted your general idea here

bonfire/config.py
bsquizz: Renamed it

bonfire/config.py
```diff
@@ -54,17 +54,33 @@
...
+    env_var = os.getenv("BONFIRE_FRONTEND_DEPENDENCIES")
+    result = set()
+
+    if env_var is None:
+        result = set(DEFAULT_FRONTEND_DEPENDENCIES)
+    elif env_var.strip() != "":
+        for app in env_var.split(","):
+            result.add(app.strip())
+
+    return result
```
Victoremepunto, line 78: is there any situation where you don't expect this to be a string ? 

Also, for non-emptiness checks, I just realized we're adding empty elements here, regardless. So, something like:
```
BONFIRE_FRONTEND_DEPENDENCIES='  ,   '
```

will have an undesirable outcome :/ 

with a comprehension list maybe:

```suggestion
    env_var = os.getenv("BONFIRE_FRONTEND_DEPENDENCIES")
    
    if env_var is None:
        return set(DEFAULT_FRONTEND_DEPENDENCIES)
    
    return set([ x for x in env_var.split(",") if x.strip()])
```

Victoremepunto (commented):
one last thing, to avoid people shooting themselves in the foot

bonfire/config.py
bsquizz: Good point, updated

Victoremepunto (approved):
LGTM!

Issue #316, opened 2023-08-01
Running "bonfire namespace list -m" shoud NOT show "warning" messages


thearifismail:
The Warning messages over-whelmed the line of interest.

Here is what I got:
```
$ bonfire namespace list -m
2023-08-01 11:30:47 [ WARNING] [          MainThread] could not retrieve reservation details for ns: ephemeral-8e7c42
2023-08-01 11:30:47 [ WARNING] [          MainThread] could not retrieve reservation details for ns: ephemeral-d23b6o
2023-08-01 11:30:47 [ WARNING] [          MainThread] could not retrieve reservation details for ns: ephemeral-h2axnp
2023-08-01 11:30:47 [ WARNING] [          MainThread] could not retrieve reservation details for ns: ephemeral-zqxeoe
NAME              RESERVED    ENV STATUS    APPS READY    REQUESTER      POOL TYPE    EXPIRES IN
----------------  ----------  ------------  ------------  -------------  -----------  ------------
ephemeral-wkywze  true        ready         9/9           thearifismail  default      9d23h32m40s
$ 
```

Would be nice to see:
```
$ bonfire namespace list -m
NAME              RESERVED    ENV STATUS    APPS READY    REQUESTER      POOL TYPE    EXPIRES IN
----------------  ----------  ------------  ------------  -------------  -----------  ------------
ephemeral-wkywze  true        ready         9/9           thearifismail  default      9d23h32m40s
$
```

bsquizz:
Fixed by #323

PR #323, opened 2023-08-28
Don't get reservation status for terminating namespaces

adamrdrew:
This patch addresses https://github.com/RedHatInsights/bonfire/issues/316 / RHCLOUD-27918

I looked at this for way too long thinking about all kinds of complex fixes (pass the logger in as a depdency so we can control its log level out at the click commands, split the list namespaces method up into mine and not mine methods to better control what we're doing at run time so we don't need to check reservation status on any !mine namespaces) but in the end I realized we should just do this, always have the log level in `Namespace.reservation` be debug. This information isn't useful to a normal bonfire user; devs or cluster admins can up the log level.

### DIFF
--- a/bonfire/namespaces.py
+++ b/bonfire/namespaces.py
@@ -226,6 +226,7 @@ class Namespace:
     def is_active(self):
         return self.phase == self.STATUS_ACTIVE
 
+
 def get_namespaces(available=False, mine=False):
     """
     Look up reservable namespaces in the cluster.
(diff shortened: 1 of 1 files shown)


adamrdrew:
Instead we should look if the phase is terminating for the namespace and then just omit it from the list. This will require we add a new is_terminating property to Namespace. Gonna do this that way instead.

bsquizz (approved):
Looks good!

PR #324, opened 2023-08-29
Handle image errors

adamrdrew:
This patch depends on https://github.com/RedHatInsights/ocviapy/pull/9

But unfortunately I can't get it to work. It looks like it should. I was testing with 

```
            "args": [
                "deploy",
                 "advisor",
                 "--set-image-tag", 
                 "quay.io/cloudservices/advisor-backend=324324",
            ],
```

in vscode but it never pops the `StatusError` and I can't figure out why.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -48,6 +48,7 @@ from bonfire.utils import (
     split_equals,
     validate_time_string,
 )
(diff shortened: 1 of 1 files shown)


adamrdrew:
Bonfire changes weren't required

PR #325, opened 2023-08-31
Use gitlab search API to find repos

bsquizz:
If there are more than 100 repos in a gitlab group, we fail to find the repo due to pagination. Thanks to @Victoremepunto for suggesting we use the search API instead.

We could potentially get over 100 results from a search, but it is highly unlikely. We can deal with the pagination at that time if it happens.


Victoremepunto (approved):
LGTM

PR #326, opened 2023-09-01
Handle status error

adamrdrew:
This patch is related to https://github.com/RedHatInsights/ocviapy/pull/9 and handles statys errors nicely


bsquizz (approved):
This is fine to go in now, causes no problems since the current version of `ocviapy` already has `StatusError`

PR #327, opened 2023-09-15
Doesn't output frontend links if no frontend is deployed

gburges:


bsquizz:
I see what you are going for, but the default login user actually applies for envs without a frontend as well. See my latest commit, will it work?

gburges:
ooh gotcha. Yes it looks alright to me

bsquizz (approved):

bsquizz (approved):

PR #329, opened 2023-10-05
Updates default iqe yaml to use env

gburges:
fixes https://github.com/RedHatInsights/bonfire/issues/328 after https://github.com/RedHatInsights/clowder/pull/855

PR #331, opened 2023-10-09
Use newly supported 'env' block in default IQE CJI template

bsquizz:
Second attempt at #329 --

* Changes the IQE CJI template to use the new `env` block introduced in https://github.com/RedHatInsights/clowder/pull/855
* Adds a `--env-var` CLI option to allow users to pass custom env vars to the IQE pod at runtime

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -335,25 +335,19 @@ def _validate_set_parameter(ctx, param, value):
         raise click.BadParameter("format must be '<component>/<param>=<value>'")
 
 
-def _validate_set_image_tag(ctx, param, value):
+def _validate_split_equals(ctx, param, value):
     try:
         return split_equals(value)
     except ValueError:
-        raise click.BadParameter("format must be '<image uri>=<tag>'")
-
-
-def _validate_set_preferred_params(ctx, param, value):
-    try:
-        return split_equals(value)
-    except ValueError:
-        raise click.BadParameter("format must be '<parameter name>=<value>'")
-
-
-def _validate_set_env_var(ctx, param, value):
-    try:
-        return split_equals(value)
-    except ValueError:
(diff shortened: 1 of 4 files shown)


bonfire/bonfire.py
```diff
@@ -884,7 +908,7 @@ def _get_apps_config(
 
     # handle git ref/image substitutions if reference environment was provided
     if ref_env:
-        apps_config = sub_refs(apps_config, ref_env, fallback_ref_env, preferred_params=prefer)
+        apps_config = sub_refs(apps_config, ref_env, fallback_ref_env, preferred_params)
```
bsquizz, line 911: Renaming 'prefer' to 'preferred_params' was supposed to go in with #333 -- not sure what happened but it was just for readability so we'll get the change in here :man_shrugging:

bonfire/bonfire.py
```diff
@@ -342,13 +342,20 @@ def _validate_set_image_tag(ctx, param, value):
...
 
-def _validate_set_prefer(ctx, param, value):
+def _validate_set_preferred_params(ctx, param, value):
     try:
         return split_equals(value)
     except ValueError:
         raise click.BadParameter("format must be '<parameter name>=<value>'")
 
 
+def _validate_set_env_var(ctx, param, value):
```
adamrdrew, line 352: Why have these two functions if they both do the same thing? Couldn't you just have a single `safe_split_equals` function?

adamrdrew (commented):
The question about those functions aside this looks good to me

bonfire/bonfire.py
bsquizz: I debated this myself for a moment too :laughing: It's because I want to have a descriptive error message for each parameter and they each have a slightly different message. I combined them all into one validator func now, take a look.

adamrdrew (approved):

adamrdrew:
Awesome thanks! This LGTM!

PR #332, opened 2023-10-11
Improve ns describe

bsquizz:
Always show gateway route (because it is applicable to access the env for backend APIs too)
Show number of ClowdApps and Frontends

### DIFF
--- a/bonfire/namespaces.py
+++ b/bonfire/namespaces.py
@@ -381,20 +381,22 @@ def describe_namespace(project_name: str):
         raise FatalError(f"namespace '{project_name}' was not reserved with namespace operator")
 
     frontends = get_json("frontend", namespace=project_name)
+    clowdapps = get_json("clowdapp", namespace=project_name)
     fe_host, keycloak_url = parse_fe_env(project_name)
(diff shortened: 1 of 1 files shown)


bonfire/namespaces.py
```diff
@@ -381,20 +381,22 @@ def describe_namespace(project_name: str):
...
         output += f"Project URL: {ns_url}\n"
     output += f"Keycloak admin route: {keycloak_url}\n"
     output += f"Keycloak admin login: {kc_creds['username']} | {kc_creds['password']}\n"
-    if frontends.get('items'):
-        output += f"Frontend route: https://{fe_host}\n"
-    else:
-        output += "No frontends deployed\n"
+    output += (
+        f"{len(clowdapps['items'])} ClowdApp(s), "
+        f"{len(frontends['items'])} Frontend(s) deployed\n"
```
Victoremepunto, line 397: isn't this at risk of raising a `KeyError` if no `items` key present in the map ?

bsquizz: Yeah, I thought about the same thing too :), just updated

PR #333, opened 2023-10-11
Add fallback ref & preference rules for duplicate deployment targets

bsquizz:
The new containerized frontend apps have two deployment targets configured in the same environment -- a 'stable' target and a 'beta' target. Because of this, bonfire doesn't know which one to select when a user indicates something like `--ref-env=insights-stage` -- the deployment config from `insights-stage` that it selects to use for the git ref/image tag to deploy can be unpredictable.

Also, when bonfire is run with no `--ref-env`, it defaults to pulling from 'main/master' for all components. This isn't ideal for the frontend deployments which are pushing their betas into the main/master branch.

This PR:
* Sets the default `--ref-env` to be 'insights-stage' (can be changed with an env var). For backends, this should have basically no impact, because nearly every stage deployment target is using `master/main` already anyways.
* Adds a '--fallback-ref-env' argument. If we can't find a component's deployment config in the ref env, we will attempt to find it in the fallback ref env (by default this is 'insights-stage', but can be changed with an env var). If both the ref env and the fallback are the same, we don't bother analyzing the fallback env.
* Adds a '--prefer' argument that allows users to customize how the bonfire "duplicate deploy target" preference logic works. For example, you can use `bonfire deploy host-inventory --frontends=true --ref-env=insights-stage --prefer ENV_NAME=frontends-beta` to prefer beta deployment targets. You can use `--prefer` multiple times if for some reason you wish to specify multiple parameters.
* Changes the logic of `_check_replace_other` to use a number weighted system, and takes the `--prefer` input into account.  Also, I noticed the 'replicas' logic needed a tweak, went ahead and updated it here.
* Sets the default for `--prefer` to be `ENV_NAME=frontends` -- which would prefer consoledot stable frontend deployments.
* Enhances the logging related to how the app/component list is built
* Fixes a bug I found related to how we match namespaces to an environment. Some namespaces have the same name across environments or clusters. So we should match on the full namespace path in app-interface instead of just the name.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -408,8 +408,8 @@ _app_source_options = [
     click.option(
         "--prefer",
         help=(
-            "Prefer deployment targets that have '<parameter name>=<value>' when duplicate"
-            f" deployment targets found in {APP_SRE_SRC}"
+            f"When there are multiple deployment targets found in {APP_SRE_SRC}, prefer the ones "
+            "that have '<parameter name>=<value>' set. Can be specified multiple times."
         ),
         multiple=True,
         default=conf.BONFIRE_DEFAULT_PREFER,
@@ -430,7 +430,16 @@ _process_options = _app_source_options + [
             "Use to set default 'ref'/'IMAGE_TAG' for apps."
         ),
         type=str,
-        default=None,
+        default=conf.BONFIRE_DEFAULT_REF_ENV,
+    ),
+    click.option(
+        "--fallback-ref-env",
+        help=(
+            f"Reference environment name in {APP_SRE_SRC} to be used if deployment configuration"
+            " not found in primary reference environment."
+        ),
+        type=str,
+        default=conf.BONFIRE_DEFAULT_FALLBACK_REF_ENV,
     ),
     click.option(
         "--set-image-tag",
@@ -832,17 +841,23 @@ def _describe_namespace(namespace):
     click.echo(describe_namespace(namespace))
 
 
-def _get_apps_config(source, target_env, ref_env, local_config_path, local_config_method, prefer):
+def _get_apps_config(
+    source, target_env, ref_env, fallback_ref_env, local_config_path, local_config_method, prefer
+):
     config = conf.load_config(local_config_path)
 
     if source == APP_SRE_SRC:
-        log.info("fetching target env apps config using source: %s, env: %s", source, target_env)
+        log.info("fetching target env apps config using source: %s", source)
         if not target_env:
             _error("target env must be supplied for source '{APP_SRE_SRC}'")
         apps_config = get_apps_for_env(target_env, preferred_params=prefer)
 
-        if target_env == conf.EPHEMERAL_ENV_NAME and not ref_env:
-            log.info("target env is 'ephemeral' with no ref env given, using 'master' for all apps")
+        if not ref_env and target_env == conf.EPHEMERAL_ENV_NAME:
+            # set git target to 'master' because ephemeral targets have no git ref defined
+            log.info(
+                "target env is '%s' and no ref env given, using 'master' git ref for all apps",
+                conf.EPHEMERAL_ENV_NAME,
+            )
             for _, app_cfg in apps_config.items():
                 for component in app_cfg.get("components", []):
                     component["ref"] = "master"
@@ -867,9 +882,9 @@ def _get_apps_config(source, target_env, ref_env, local_config_path, local_confi
                 # re-raise with a bit more context
                 raise FatalError(f"{str(err)}, hit on app {app_name}")
 
+    # handle git ref/image substitutions if reference environment was provided
     if ref_env:
-        log.info("subbing app template refs/image tags using env: %s", ref_env)
-        apps_config = sub_refs(apps_config, ref_env, preferred_params=prefer)
+        apps_config = sub_refs(apps_config, ref_env, fallback_ref_env, preferred_params=prefer)
 
     return apps_config
 
@@ -893,6 +908,7 @@ def _process(
     local_config_method,
     set_image_tag,
     ref_env,
+    fallback_ref_env,
     target_env,
     set_template_ref,
     set_parameter,
@@ -909,7 +925,13 @@ def _process(
     prefer,
 ):
     apps_config = _get_apps_config(
-        source, target_env, ref_env, local_config_path, local_config_method, prefer
+        source,
+        target_env,
+        ref_env,
+        fallback_ref_env,
+        local_config_path,
+        local_config_method,
+        prefer,
     )
 
     processor = TemplateProcessor(
@@ -955,6 +977,7 @@ def _cmd_process(
     local_config_method,
     set_image_tag,
     ref_env,
+    fallback_ref_env,
     target_env,
     set_template_ref,
     set_parameter,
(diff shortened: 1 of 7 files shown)


bonfire/qontract.py
```diff
@@ -146,19 +146,28 @@ def _find_matching_component(apps, app_name, component_name):
 
 
 def _check_replace_other(other_params, this_params):
+    # prefer stable frontend targets over beta targets
+    this_frontend_stable = this_params.get("ENV_NAME") != "frontends-beta"
+    other_frontend_stable = other_params.get("ENV_NAME") != "frontends-beta"
+    if this_frontend_stable and not other_frontend_stable:
+        return True
```
adamrdrew, line 153: Negative conditionals and double negatives confuse me, could we recast this like:

```
def _is_stable_frontend(params):
    return params.get("ENV_NAME") == "frontends"
    
def _check_replace_other(other_params, this_params):
    this_frontend_stable = _is_stable_frontend(this_params)
    other_frontend_stable = _is_stable_frontend(other_params)
    
    if this_frontend_stable and not other_frontend_stable:
        return True
```

bsquizz: Check it out now, renamed it to `def _is_preferred(params)`

bsquizz: I think it opens up the door to being able to customize what is preferred or not in future. Like for example we could have a CLI argument for `--prefer ENV_NAME=frontends-beta` ... we could even allow you to prefer multiple parameters which are OR'd or AND'd together.

bonfire/qontract.py
```diff
@@ -396,26 +374,3 @@ def sub_refs(apps, ref_env_name, preferred_params):
                 final_apps[app_name]["components"][idx]["ref"] = "master"
 
     return final_apps
-
-
-def get_namespaces_for_env(environment_name):
```
bsquizz, line 356: No longer used

bonfire/qontract.py
```diff
@@ -118,11 +104,6 @@ def get_env(self, env):
     def get_apps(self):
         return self.client.execute(APPS_QUERY)["apps"]
 
-    def get_namespace(self, name):
```
bsquizz, line 121: Tech debt, no longer used

bonfire/qontract.py
```diff
@@ -66,22 +68,6 @@
     """
 )
 
-NAMESPACE_QUERY = gql(
```
bsquizz, line 69: Tech debt, no longer used

bonfire/bonfire.py
```diff
@@ -398,6 +405,16 @@ def _validate_opposing_opts(ctx, param, value):
         default=conf.EPHEMERAL_ENV_NAME,
         show_default=True,
     ),
+    click.option(
+        "--prefer",
+        help=(
+            "Prefer deployment targets that have '<parameter name>=<value>' when duplicate"
```
adamrdrew, line 411: This is worded funny. I'm not sure I'd understand it if I saw it IRL. Maybe something like

```
"When there are multiple deployment targets prefer the one that has '<parameter name>=<value>' set. Can be specified multiple times."
```

I know its wordier but maybe?

bonfire/bonfire.py
```diff
@@ -815,14 +832,14 @@ def _describe_namespace(namespace):
     click.echo(describe_namespace(namespace))
 
 
-def _get_apps_config(source, target_env, ref_env, local_config_path, local_config_method):
+def _get_apps_config(source, target_env, ref_env, local_config_path, local_config_method, prefer):
```
adamrdrew, line 835: maybe changed `prefer` to `preferred_params` here to retain consistency? Also, `prefer` gives me "this is probably a boolean" vibes whereas `preferred_params` sounds more like some kind of collection or dictionary.

bonfire/config.py
```diff
@@ -45,7 +45,10 @@
 QONTRACT_PASSWORD = os.getenv("QONTRACT_PASSWORD", APP_INTERFACE_PASSWORD or None)
 QONTRACT_TOKEN = os.getenv("QONTRACT_TOKEN")
 
-BASE_NAMESPACE_NAME = os.getenv("BASE_NAMESPACE_NAME", "ephemeral-base")
+BASE_NAMESPACE_PATH = os.getenv(
+    "BASE_NAMESPACE_PATH",
+    "/services/insights/ephemeral/namespaces/ephemeral-base.yml",
```
adamrdrew, line 50: Why did you need to change this?

adamrdrew (approved):
LGTM! I was expecting this patch to be more intense but it was really less invasive than I pictured. I guess in the end this was just taking what was already there and just adding a new way to narrow down the querying. So, LGTM!

bonfire/config.py
bsquizz: This was related to: 

> I also found a bug related to how we match namespaces to an environment. Some namespaces now have the same name across environments. So now, we match on the full namespace path instead of just the name.

bsquizz: As an example, `frontends` and `frontends-base` exist in both stage and prod envs. But even for ephemeral-base, if you look for all namespaces named 'ephemeral-base' in app-interface configs, you'll actually find there's more than 1 :) There is one tied to the ephemeral cluster, and one tied to the dev cluster.

bonfire/bonfire.py
bsquizz: Good idea, will implement

bonfire/bonfire.py
bsquizz: I like it, will change

bsquizz:
Tested this to ensure there's no impact to the backend configs. Here's a little script I'm using to ensure that every resource between 2 `bonfire process` commands is equivalent:
```python
# compare.py <old file> <new file>

import json
import sys


with open(sys.argv[1]) as fp:
    old = json.loads(fp.read())

with open(sys.argv[2]) as fp:
    new = json.loads(fp.read())

for old_item in old["items"]:
    kind = old_item["kind"]
    name = old_item["metadata"]["name"]

    for new_item in new["items"]:
        if new_item["kind"] == kind and new_item["metadata"]["name"] == name:
            if old_item != new_item:
                print(f"diff found: {kind} {name}")
```

I am switching between bonfire's 'master' branch, for ex:
```
git checkout master
bonfire process patchman --frontends=true --clowd-env=test > old-patchman
git checkout prefer_stable_frontend_target
bonfire process patchman --frontends=true --clowd-env=test > new-patchman
python compare.py old-patchman new-patchman
```

Here are results for `advisor`:
```
❯ python compare.py old-advisor new-advisor
diff found: Frontend chrome
diff found: Frontend rbac
diff found: Frontend landing
diff found: Frontend dashboard
diff found: Frontend inventory
diff found: Frontend advisor
diff found: Secret sources-api-secrets
```

(the secret is different every time because the template is generating a random pre-shared key)

For patchman:
```
❯ python compare.py old-patchman new-patchman
diff found: Frontend patch
diff found: Frontend dashboard
diff found: Frontend rbac
diff found: Frontend chrome
diff found: Frontend inventory
diff found: Frontend landing
```

(the frontends are all different because now we're selecting the stage-stable frontends)

When using `--ref-env=insights-production`:
```
❯ python compare.py old-advisor new-advisor
diff found: Secret sources-api-secrets
❯ python compare.py old-patchman new-patchman
diff found: Frontend patch
```

(the patch frontend is different because we're falling back to select the stage-stable frontend... previously bonfire would have selected the stage-beta frontend)

bonfire/config.py
adamrdrew: OK that makes sense! Thanks for the clarification.

adamrdrew (approved):
LGTM - I don't see any problems or anything that should change. All makes sense.

PR #336, opened 2023-11-17
Add handling for bonfire.dependencies annotation

mjholder:
ticket: https://issues.redhat.com/browse/RHCLOUD-28382

This is adding logic for bonfire to gather dependencies listed under the annotation `bonfire.dependencies`. This supports the need for some apps to have dependencies that are not clowdapps.

### DIFF
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -383,7 +383,7 @@ def get_dependencies(items, optional=False):
     clowdapp_items = [item for item in items if item.get("kind").lower() == "clowdapp"]
 
     deps_for_app = dict()
-    
+
     for clowdapp in clowdapp_items:
         name = clowdapp["metadata"]["name"]
         dependencies = {d for d in clowdapp["spec"].get(key, [])}
--- a/bonfire/utils.py
+++ b/bonfire/utils.py
@@ -383,12 +383,18 @@ def get_dependencies(items, optional=False):
     clowdapp_items = [item for item in items if item.get("kind").lower() == "clowdapp"]
 
     deps_for_app = dict()
-
+    
     for clowdapp in clowdapp_items:
         name = clowdapp["metadata"]["name"]
         dependencies = {d for d in clowdapp["spec"].get(key, [])}
         log.debug("clowdapp '%s' has %s: %s", name, key, list(dependencies))
         deps_for_app[name] = dependencies
+
+    for item in items:
+        name = item.get("metadata", {}).get("name")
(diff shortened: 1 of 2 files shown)


bonfire/utils.py
```diff
@@ -390,6 +390,14 @@ def get_dependencies(items, optional=False):
         log.debug("clowdapp '%s' has %s: %s", name, key, list(dependencies))
         deps_for_app[name] = dependencies
 
+    for item in items:
+        name = item.get("metadata", {}).get("name")
+        bonfire_deps = item.get("metadata", {}).get("annotations", {}) \
+            .get("bonfire.dependencies", "").split(',')
+        filtered_bonfire_deps = list(filter(lambda x: x != '', bonfire_deps))
```
adamrdrew, line 397: What you have is fine and works but you could be a bit more idiomatic by using a list comprehension

```
filtered_bonfire_deps = [dep for dep in bonfire_deps if dep]
```

bonfire/utils.py
```diff
@@ -390,6 +390,14 @@ def get_dependencies(items, optional=False):
         log.debug("clowdapp '%s' has %s: %s", name, key, list(dependencies))
         deps_for_app[name] = dependencies
 
+    for item in items:
+        name = item.get("metadata", {}).get("name")
+        bonfire_deps = item.get("metadata", {}).get("annotations", {}) \
```
adamrdrew, line 395: You get the metadata multiple times, so you could slim this down a bit by storing that

```
    metadata = item.get("metadata", {})
    name = metadata.get("name")
    bonfire_deps = metadata.get("annotations", {}).get("bonfire.dependencies", "").split(',')
```

Its OK the way it is but food for thought

adamrdrew (approved):
LGTM!!!

bsquizz (changes requested):
+1 on Adam's suggestions  ... also, we need to link up pre-commit to this repo but can you run this once you make your changes, it will probably automatically adjust your multi-line formatting on line 395:
```
black **/*.py --line-length=100
```

You might need to `pip install` that if you don't have `black` yet.

bonfire/utils.py
mjholder: that's a very good point. I'll update it

bonfire/utils.py
mjholder: i did notice there was more list comprehensions than lambda's present. I'll make those changes so it fits in better

mjholder (commented):
I linted the changes with black

bsquizz (approved):

PR #339, opened 2023-12-06
Telemetry support

mjholder:
This PR adds telemetry gathering which sends the data to an Elastic instance. 
The new logger that sends the data can fail without effecting the performance of bonfire.

Right now I have it only enabled if it is called by the bonfire bot.
Users can opt-out by setting the env variable 'ENABLE_TELEMETRY=false'


bonfire/elastic_logging.py
```diff
@@ -0,0 +1,30 @@
...
+
+class AsyncElasticsearchHandler(logging.Handler):
+    def __init__(self, es_url):
+        super().__init__()
+        self.es_url = es_url
+        self.executor = ThreadPoolExecutor(max_workers=10)
+
+    def emit(self, record):
+        log_entry = self.format(record)
+        if conf.ENABLE_TELEMETRY == 'true' and conf.BONFIRE_BOT:
```
bsquizz, line 16: If we make it a bool, this line can be changed to:
```suggestion
        if conf.ENABLE_TELEMETRY:
```

If you are trying to use the bot check to determine whether telemetry is "on" or "off", honestly I think we can just use 1 env var here. We'll just set up the jenkins jobs to pass `ENABLE_TELEMTRY=true` as an env var and that can take care of it.

bonfire/elastic_logging.py
```diff
@@ -0,0 +1,30 @@
...
+        # Convert log_entry to JSON and send to Elasticsearch
+        try:
+            headers = {"Authorization": conf.ELASTICSEARCH_APIKEY,
+                       "Content-Type": "application/json"}
+
+            response = requests.post(self.es_url, headers=headers, data=log_entry, timeout=0.1)
+            response.raise_for_status()
+        except Exception as e:
+            # Handle exceptions (e.g., network issues, Elasticsearch down)
+            print(f"Error sending data to elasticsearch: {e}")
```
bsquizz, line 29: Instead of `print` can we use `log.error` here (or `log.exception` if you want to print the whole traceback)

```suggestion
            log.error("Error sending data to elasticsearch: %s", e)
```

bonfire/elastic_logging.py
```diff
@@ -0,0 +1,30 @@
...
+            pass
```
bsquizz, line 30: `pass` can be removed, it is redundant now since you have a print/log statement on the line above.

bonfire/bonfire.py
```diff
@@ -51,7 +54,12 @@
     merge_app_configs,
 )
 
+
 log = logging.getLogger(__name__)
+es_telemetry = logging.getLogger("elasticsearch")
```
bsquizz, line 59: Right now if we import anything from `bonfire` it would set up this logger. So I think even during a unit test (for example), this logger is going to get set up. I am thinking we should move the setup of this into a function... like `def setup_telemetry` and maybe you have the `main_with_handler` func call `setup_telemetry()`?

bonfire/bonfire.py
```diff
@@ -81,6 +89,51 @@ def current_namespace_or_error():
     return namespace
 
 
+def option_usage_wrapper(command):
```
bsquizz, line 92: I'm still thinking we could capture the commands in 1 place more easily... do you need the wrapper? For example if we want to just record the command that was executed to start with, what if we just record the value of `sys.argv` when we are in the `main_with_handler` function?

For example you can see if you add this code:

```
def main_with_handler():
    print(sys.argv[1:])
    try:
        main()
    except FatalError as err:
        _error(str(err))
```

It results in this:
```
❯ bonfire process foo --set-parameter someapp/someparam=somevalue --clowd-env=myenv
['process', 'foo', '--set-parameter', 'someapp/someparam=somevalue', '--clowd-env=myenv']
```

bonfire/bonfire.py
```diff
@@ -81,6 +89,51 @@ def current_namespace_or_error():
...
+                elif arg == '-p' or arg == '--set-parameter':
+                    is_parameter = True
+
+            start_time = time.time()
+
+            log_entry = {
+                "timestamp": datetime.datetime.now().isoformat(),
+                "duration_seconds": time.time() - start_time,
+                "bot": conf.BONFIRE_BOT,
+                "command": command,
```
bsquizz, line 112: What if we move some of this data right into the ` AsyncElasticsearchHandler` class?

When we initialize the instance at the startup of bonfire (in `main_with_handler`), we can give that instance some a 'metadata' variable. So for example, the class could look like this:

```
class AsyncElasticsearchHandler(logging.Handler):
    def __init__(self, es_url, metadata):
        super().__init__()
        self.es_url = es_url
        self.executor = ThreadPoolExecutor(max_workers=10)
        self.metadata = metadata
```

and in `main_with_handler` we could do something like:
```python
metadata = {
    "uuid": uuid.uuid4()
    "start_time": datetime.datetime.now().isoformat(),
    "bot": conf.BONFIRE_BOT,
    "command": _mask_parameter_values(sys.argv[1:]),  # need to write code for  _mask_parameter_values
}
es_handler = AsyncElasticsearchHandler(conf.ELASTICSEARCH_HOST, metadata)
```

and every time that `emit` is called, it could add this `metadata` into the JSON data that it sends? This could be considered the `run metadata` or something that never changes for the entire `bonfire` run until the process exits. As long as the `uuid` stays the same for the whole process execution -- it would allow us to track things for the same run of bonfire.

You may be able to have some helper funcs in the class itself that add on the `success` or the `failure` data as well.

Basically the end goal here is: I think we want to be able to call `send_telementry("something happened...")` at more parts of the code in future... kind of like how we might call `log.info("something happened...") at various places. So I am thinking we want some helper functions defined that can be called in various pieces of code. When we call them, the "run ID" that is sent to ES should stay the same ... even though we might send a few telemetry messages for 1 bonfire command execution.

bsquizz (changes requested):
Good start I think! I have a few suggestions and a couple ideas to think on.

bonfire/bonfire.py
mjholder: I moved the handler setup to `main_with_handler`. It does make some things a little more complicated. i.e. Grabbing the handler from the logger because the handler isn't in scope. Not a big deal though

bonfire/bonfire.py
mjholder: I added a function to bonfire, `send_telemetry`  that we can call anywhere. The parameter values get replaced with `******`

bonfire/bonfire.py
mjholder: I added a couple helper functions to get this working the way you describe. I'm not exactly sure where these helper functions should live. Right now I have them in `bonfire.py` but I may move them to `elastic_logging.py` to keep all the related functions together. 
The only reason I didn't do that yet is because it will make things a bit more complex. ie. making sure a logger exists. If I design it so the handler stores a reference to the logger and just call the logger from the handler, well that's just weird I guess. Having the handler actually be the logger.
There may be a better solution, I'm just still brainstorming on what that is

mjholder: Ok I moved some pieces around so that the logic for elasticsearch is only in `elastic_logging.py`. Now you would just create an instance of `ElasticLogger` at the top of a file and call the object's `send_telemetry()`.

bsquizz (changes requested):
I like where this is going! A couple minor tweaks to suggest

bonfire/elastic_logging.py
```diff
@@ -0,0 +1,82 @@
...
+class AsyncElasticsearchHandler(logging.Handler):
+    def __init__(self, es_url, metadata):
+        super().__init__()
+        self.es_url = es_url
+        self.metadata = metadata
+        self.executor = ThreadPoolExecutor(max_workers=10)
+
+    def emit(self, record):
+        self.metadata["time_of_logging"] = dt.now().isoformat()
+        self.metadata["duration"] = str(dt.now() - dt.fromisoformat(self.metadata["start_time"]))
```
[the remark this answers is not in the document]
mjholder: ok I did that. The only caveat is that I need to format the start_time before I pass the metadata to the request. I add a small piece that makes a copy of the metadata that's formatted.
I didn't want to modify the original metadata just in case there would be multiple logs each run.

bsquizz (approved):

PR #343, opened 2024-01-17
None is not respected for namespace as clowd_env gets in the way

skarekrow:
Currently when you don't specfiy namespace for `process` you get an error that you need to supply one. This is incorrect, as you could simply say the namespace is `foo` or even `None` and it will proceed correctly.

Just skipping `_get_env_name` since there is no environment fixes the issue.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1059,7 +1059,10 @@ def _cmd_process(
     preferred_params,
 ):
     """Fetch and process application templates"""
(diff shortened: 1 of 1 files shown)


bsquizz:
OK I see what you are going for here... there were a few changes to make in a few areas -- closing in favor of https://github.com/RedHatInsights/bonfire/pull/345 -- can you let me know what you think of that approach?

PR #344, opened 2024-01-18
Add env vars to specify trusted apps/components

bsquizz:
If an app or component is found in these lists, then we will not strip the resource configs off of its template during processing.

Using a hard-coded list like this is a work-around until we fully develop the way to "intelligently trust" app resource configurations.

The host-inventory app has been audited and resource configs will be adjusted via https://github.com/RedHatInsights/insights-host-inventory/pull/1595


maknop:
Simple enough! Validates that the app or component isn't in the trusted list and removes it. LGTM

maknop (approved):

PR #345, opened 2024-01-18
Improved get_env_name logic, allows running `process` cmd without `--namespace` nor `--clowd-env`

bsquizz:
Currently we check that a user either passed `--namespace` or `--clowd-env` when running the `process` command to prevent a user from processing templates with an invalid ClowdEnvironment name. This is because we always set the `ENV_NAME` parameter when we process templates (https://github.com/RedHatInsights/bonfire/blob/master/bonfire/processor.py#L577) due to the parameter being required on ClowdApp templates.

But this makes running 'process' for dry run/offline use a bit annoying since you have to always specify a fake argument for one of those options. And also, if you are not processing a template with ClowdApps, there is not much need to supply this argument.

The `deploy` command already has safeguards in it which check to make sure the ClowdEnvironment can be located. I think we can assume that if a user is using `process` without any args, they are doing it for a reason, and we can just log some informative warnings. Worst case scenario is they apply the config with the ENV_NAME set to `None`. In case they do, I have changed the way that `wait_for_all_resources` runs to catch the scenario where a ClowdApp exists but it has a bad ClowdEnvironment name set.

This PR changes the logic for `get_env_name`:
1. If `--clowd-env` is given, just use that.
2. If `--namespace` is given, try to find the ClowdEnvironment tied to it -- if we can't, log a warning, ENV_NAME will be `None`
3. If neither params are given, ENV_NAME will be `None`

And now the `_all_resources_ready` check now determines which ClowdEnvironments to wait on based on what it sees set in ClowdApps. For example if one is in the namespace with a value of `None`, the user now sees this:
```
2024-01-19 12:38:24 [    INFO] [          MainThread] will wait on ClowdEnvironment 'None' found on ClowdApp .spec.envName
2024-01-19 12:38:24 [    INFO] [thread-5 (wait_for_ready)] [clowdenvironment/none] waiting up to 600sec for resource to be 'ready'
```

Since get_env_name used to be shared by the `process-env` / `deploy-env` commands, to simplify things they will no longer use that function and simply require `-n` to specify the ClowdEnvironment target namespace. This is not a commonly used command and the only place in our docs (related to local setup on minikube) where we did mention it used `-n` already.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -601,8 +601,9 @@ _clowdenv_process_options = [
     click.option(
         "--namespace",
         "-n",
-        help="Target namespace of the ClowdEnvironment (default: none)",
+        help="Target namespace of the ClowdEnvironment",
         type=str,
+        required=True,
     ),
     click.option(
         "--quay-user",
@@ -613,9 +614,7 @@ _clowdenv_process_options = [
     click.option(
         "--clowd-env",
         "-e",
-        help=(
-            "Name of ClowdEnvironment (default: if --namespace provided, will try to find match)"
-        ),
+        help=("Name of ClowdEnvironment (default: env-<namespace>)"),
         type=str,
         default=None,
     ),
@@ -1363,9 +1362,10 @@ def _cmd_config_deploy(
         click.echo(ns)
 
 
-def _process_clowdenv(target_namespace, quay_user, env_name, template_file, local):
-    env_name = _get_env_name(target_namespace, env_name)
-    return process_clowd_env(target_namespace, quay_user, env_name, template_file, local)
+def _process_clowdenv(namespace, quay_user, clowd_env, template_file, local):
+    if not clowd_env:
+        clowd_env = f"env-{namespace}"
+    return process_clowd_env(namespace, quay_user, clowd_env, template_file, local)
 
 
 @main.command("process-env")
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -960,8 +960,8 @@ def _get_env_name(ns, env_name):
             log.info("searching for ClowdEnvironment tied to ns '%s'...", ns)
             match = find_clowd_env_for_ns(ns)
             if not match:
-                _error(
-                    f"could not find a ClowdEnvironment with target ns '{ns}'.  "
+                log.warning(
+                    "could not find a ClowdEnvironment with target ns '%s'.  "
                     "Specify one with '--clowd-env' if needed."
                 )
             else:
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -516,7 +516,7 @@ _process_options = _app_source_options + [
         "--clowd-env",
         "-e",
         help=(
-            f"Name of ClowdEnvironment (default: if --namespace provided, {conf.ENV_NAME_FORMAT})"
+            "Name of ClowdEnvironment (default: if --namespace provided, will try to find match)"
         ),
         type=str,
         default=None,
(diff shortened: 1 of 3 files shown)


bonfire/bonfire.py
```diff
@@ -953,14 +954,22 @@ def _get_apps_config(
...
+                log.warning(
+                    "could not find a ClowdEnvironment with target ns '%s'.  "
+                    "Specify one with '--clowd-env' if needed."
+                )
+            else:
+                env_name = match["metadata"]["name"]
+        else:
+            log.warning("neither '--clowd-env' nor '--namespace' provided")
+
+    log.info("templates will be processed with parameter ENV_NAME='%s'", env_name)
```
adamrdrew, line 972: This is good, decomposing this code out into a method to be used elsewhere. But, could I propose expressing this differently?

```
def _log_get_env_name(env_name):
    log.info("templates will be processed with parameter ENV_NAME='%s'", env_name)
    return env_name

def _get_env_name(ns, env_name):
    if env_name:
        return _log_get_env_name(env_name)

    if not ns:
        log.warning("neither '--clowd-env' nor '--namespace' provided")
        return _log_get_env_name(None)
    
    log.info("searching for ClowdEnvironment tied to ns '%s'...", ns)
    match = find_clowd_env_for_ns(ns)
    if not match:
        log.warning(
            "could not find a ClowdEnvironment with target ns '%s'.  "
            "Specify one with '--clowd-env' if needed."
        )
        return _log_get_env_name(None)

    return _log_get_env_name(match["metadata"]["name"])
```

The suggestion isn't merely stylistic, or that I think its better aesthetically. When I reviewed this method I had to trace through each one of the nested ifs to try and understand every path we could take through the method. I found it hard to do, and I know if I had to work on this method later I'd have to spend a while figuring out what all are the paths through here and what's being set where. The proposed version is flat, has no nested logic, and we always return early. I add the log method so we still get the DRY log message you were going for without needing to let all execution paths dump into the return.

bonfire/openshift.py
```diff
@@ -85,21 +85,35 @@ def _resources_for_ns_wait():
...
 
-    # wait on ClowdEnvironment, if there's one using this ns as its targetNamespace
+    # wait on ClowdEnvironments if any ClowdApps reference one
     start = time.time()
 
-    clowd_env = find_clowd_env_for_ns(namespace)
-    if clowd_env:
+    clowd_envs = []
+
+    clowdapps = get_json("clowdapp", namespace=namespace)
```
adamrdrew, line 93: You could simplify this a bit if clowd_envs was a set because then you can just use add and it wont add duplicates. I wouldn't die on this hill, but wanted to throw it out there. I also made getting stuff off the clowdapps json a bit safer, I don't know if there's any danger there

```
clowd_envs = set()

clowdapps = get_json("clowdapp", namespace=namespace)
for clowdapp in clowdapps.get("items", []):
    env_name = clowdapp.get("spec", {}).get("envName")
    if env_name:
        log.info("will wait on ClowdEnvironment '%s' found on ClowdApp .spec.envName", env_name)
        clowd_envs.add(env_name)
```

adamrdrew (approved):
LGTM - the core of this change is a nice reuse of code that was already there. I made some suggestions for safety and to make it easier to understand but they are optional. approved as is.

bonfire/openshift.py
bsquizz: You're right, I had it as a 'set' at first. But then, I want to check if it is already in the set in order to only log the message once and I went back to a list with it. I can make it a set, no problem.

bonfire/bonfire.py
bsquizz: I like it, clearer logic.

PR #348, opened 2024-01-24
Fix bug in app_or_component_selector callback

### LINKED ISSUE #347: `all` seems not to be valid to prevent all resources from being stripped out of the templates
Seems like 5.4.0 introduced a bug when using the `--no-remove-resources all` parameter.

In the previous stable version (5.3.1) the templates seem to keep their resources intact (tested with `tower-analytics` and `process`) 

You can compare the resulting template on both 5.3.1 and 5.4.0 or 5.5 , one keeps the resources limits and requests while the other doesn't. 

I'm unsure if this is a bug, please @bsquizz  can you confirm this is not intentional?

bsquizz:
Fixes #347 

If a user passes `--no-remove-resources all` then both `no_remove_resources.select_all` and `remove_resources.select_all` get set to `True`. This causes `remove_option.empty` to evaluate to `False` which then causes `remove_for_none_no_exceptions` (in `bonfire.processor._should_remove` function) to not have the expected value of `True`


tests/test_bonfire.py
```diff
@@ -347,3 +347,31 @@ def test_describe_wrong_ns(mocker):
...
+
+    runner = CliRunner()
+    result = runner.invoke(bonfire.test, ["process", "some-app"])
+    assert result.exit_code == 0
+
+    kwargs = get_return_args.call_args.args[1]
+    assert kwargs["remove_resources"].select_all is True
+    assert kwargs["no_remove_resources"].select_all is False
+    assert kwargs["remove_dependencies"].select_all is False
+    assert kwargs["no_remove_dependencies"].select_all is True
```
Victoremepunto, line 363: I like that you added a couple of tests with the feature, so kudos to that :+1: 

But - I think this is testing an internal implementation detail (like, check if some values are set to true). 

If the implementation changes this test is in need to change too, besides, the output could still have the resources stripped in case there's another bug and this test would still pass right ?

I'd say we need a test that processes a template and that asserts on the processed output. 

Again, thanks for adding some tests along the fix!!

Victoremepunto (approved):
This seems to fix  #348

tests/test_bonfire.py
bsquizz: We are testing this in 2 parts, and I think when you combine the 2 parts you end up with the more "end to end" type test you are looking for. This set of tests covers that the options for `remove_resourcces` / `no_remove_resources` get set as expected when the CLI options are provided... then we cover if the `_should_remove` function behaves as expected and we test a wide range of combinations of `remove_resources` / `no_remove_resources` here: https://github.com/RedHatInsights/bonfire/blob/master/tests/test_processor.py#L405-L590

PR #350, opened 2024-01-30
ES Telemetry status logs

mjholder:
Add logs for when data transmission starts and ends successfully


bsquizz (approved):
May need to dial the logging back once we start sending more messages but for now since we send so few, this is ok

PR #352, opened 2024-02-05
Add dockerfile for Bonfire

Victoremepunto:
This adds support to build a Docker image of Bonfire with support for Openshift credentials via environment variables or Kube config files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ podman run -it --rm -e OC_LOGIN_SERVER=https://api.openshift.example.com:6443 \
 It can also make use of a Kube config file by mounting it on the container's HOME:
 
 ```
-podman run -it --rm -v $HOME/.kube/config:/opt/bonfire/.kube/config:Z,U \
+podman run -it --rm -v ${KUBECONFIG:="$HOME/.kube/config"}:/opt/bonfire/.kube/config:Z,U \
     quay.io/cloudservices/bonfire namespace list
 ```
 
--- a/README.md
+++ b/README.md
@@ -67,13 +67,13 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
 
 **Please note**: When using the Github token (classic), it **must have** the `read:project` scope for the template fetching to work!
 
-## Using Docker
+## Using a container image (podman/docker)
 
-We provide a Docker image at quay.io/cloudservices/bonfire. The image wraps a Python virtual environment with **crc-bonfire** installed on it.
+We provide a container image at quay.io/cloudservices/bonfire. The image wraps a Python virtual environment with **crc-bonfire** installed on it.
 
-In order to use Bonfire you have to provide either a valid KUBE_CONFIG or a URL and a TOKEN to connect to the target Openshift cluster.
+In order to use Bonfire you have to provide either a valid KUBECONFIG or a URL and a TOKEN to connect to the target OpenShift cluster.
 
-The container's entrypoint already invokes the **bonfire** CLI with the arguments passed to the container, so you should be able to run any bonfire command you'd typically run by using:
+The container's entrypoint already invokes the **bonfire** CLI with the arguments passed to the container, so you should be able to run any bonfire command you'd typically run, for example:
 
 ```
 podman run -it --rm quay.io/cloudservices/bonfire version
@@ -89,17 +89,17 @@ podman run -it --rm -e OC_LOGIN_SERVER=https://api.openshift.example.com:6443 \
     quay.io/cloudservices/bonfire namespace list
 ```
 
-It can also make use of a Kube config file by mounting it on the container's HOME:
+You can also make use of the kubeconfig file on your host by mounting it into the container's `$HOME` directory:
 
 ```
-podman run -it --rm -v ${KUBECONFIG:="$HOME/.kube/config"}:/opt/bonfire/.kube/config:Z,U \
+podman run -it --rm --userns=keep-id:uid=1000,gid=1000 -v ${KUBECONFIG:="$HOME/.kube/config"}:/opt/bonfire/.kube/config:Z \
     quay.io/cloudservices/bonfire namespace list
 ```
 
-Although you can define the KUBECONFIG location using a variable:
+If you desire for the kubeconfig to live in a non-standard location in the container, you can mount the kubeconfig into another path in the container, and set the KUBECONFIG environment variable:
 
 ```
-podman run -it --rm -v $HOME/.kube/config:/opt/kubeconfig:Z,U \
+podman run -it --rm --userns=keep-id:uid=1000,gid=1000 -v $HOME/.kube/config:/opt/kubeconfig:Z \
     -e KUBECONFIG=/opt/kubeconfig \
     quay.io/cloudservices/bonfire namespace list
 ```
--- a/README.md
+++ b/README.md
@@ -76,15 +76,32 @@ In order to use Bonfire you have to provide either a valid KUBE_CONFIG or a URL 
 The container's entrypoint already invokes the **bonfire** CLI with the arguments passed to the container, so you should be able to run any bonfire command you'd typically run by using:
 
 ```
-podman run -it --rm quay.io/cloudservices/bonfire --version
+podman run -it --rm quay.io/cloudservices/bonfire version
 ```
 
 ### Using credentials
 
-The Bonfire image expects the OC_LOGIN_SERVER and OC_LOGIN_TOKEN variables to be defined in order to connect to an Openshift cluster.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+podman run -it --rm -e OC_LOGIN_SERVER=https://api.openshift.example.com:6443 \
+    -e OC_LOGIN_TOKEN=some-api-token \
+    quay.io/cloudservices/bonfire namespace list
+```
+
+It can also make use of a Kube config file by mounting it on the container's HOME:
+
+```
+podman run -it --rm -v $HOME/.kube/config:/opt/bonfire/.kube/config:Z,U \
+    quay.io/cloudservices/bonfire namespace list
```
bsquizz, line 96: After some reading, I think this is how we can run without messing up the local kube config permissions:
```
--userns=keep-id:uid=1000,gid=1000
```
(using 1000 since that is the uid and gid of the user you set up in your Dockerfile)

Seems to be working for me. Read this GH issue and the last comment [here](https://github.com/containers/podman/issues/7052) -- the option is now documented [here](https://docs.podman.io/en/v4.4/markdown/options/userns.container.html#userns-mode) 

The other thing that works for me is running with `-u root`. If you do that, you need to add `ENV KUBECONFIG=/opt/bonfire/.kube/config` to the Dockerfile. [This article](https://www.tutorialworks.com/podman-rootless-volumes/ ) was helpful explaining the user namespaces, but it didn't mention the `--userns` options, those are probably newer :)

entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+
+init_check() {
+
+    if check_kube_config; then
+        return 0
+    fi
+
+    if openshift_credentials_present; then
+        if ! try_login_openshift; then
+            echo "Failed logging into Openshift!"
```
bsquizz, line 11:
```suggestion
            echo "Failed to log into OpenShift!"
```

bsquizz (changes requested):
Looks good to me, let's just figure out how we want to recommend running the container for both:
1. A CI environment (jenkins/tekton/etc.)
2. A local user environment

Did some research and found some options, see the comments :)

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+```
+
+Although you can define the KUBECONFIG location using a variable:
+
+```
+podman run -it --rm -v $HOME/.kube/config:/opt/kubeconfig:Z,U \
```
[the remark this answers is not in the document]
Victoremepunto: I agree, I was also considering skipping this part .... I think we can definitely not include this.

README.md
Victoremepunto: I think running with root is what happens when you use regular Docker or force root on podman. I want to believe rootless podman is where we need to strive to whenever possible. 

Thanks for sharing that Podman Issue - that was revealing, it was a great finding. So reading it, it basically confirms on Docker or running Podman with root you get a consistent behavior, but if you try running rootless podman you face this issue (and someone in the thread points out you'll face the same too if running rootless Docker). 

So it looks like, specially considering Nov 22' entry mentions this new feature, all we have to do when using Podman is to use the `--userns=keep-id` option. 

I tested it and works like a charm, so I think you found the solution ! Awesome ! :rocket: 

My only concern is that is "relatively new" feature in podman. I'm on Fedora 39, podman version I have is 4.8.3 

For the moment, I'm gonna support using this since:
- It works like a charm in rootless podman
- It allows to suggest a way to share the Kube config context with the container, which makes changes to the context that persist out of the container without messing the file's permissions (so subsequent local `oc` or `kubectl` commands should be okay).

I like this. Thanks again, great job Brandon !

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+
+The Bonfire image can use the `OC_LOGIN_SERVER` and `OC_LOGIN_TOKEN` variables to connect to an Openshift cluster.
+
+```
+podman run -it --rm -e OC_LOGIN_SERVER=https://api.openshift.example.com:6443 \
+    -e OC_LOGIN_TOKEN=some-api-token \
+    quay.io/cloudservices/bonfire namespace list
+```
+
+It can also make use of a Kube config file by mounting it on the container's HOME:
```
bsquizz, line 92:
```suggestion
You can also make use of the kubeconfig file on your host by mounting it into the container's `$HOME` directory:
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+
+```
+podman run -it --rm -v ${KUBECONFIG:="$HOME/.kube/config"}:/opt/bonfire/.kube/config:Z,U \
+    quay.io/cloudservices/bonfire namespace list
+```
+
+Although you can define the KUBECONFIG location using a variable:
```
bsquizz, line 99:
```suggestion
If you desire for the kubeconfig to live in a non-standard location in the container, you can mount the kubeconfig into another path in the container, and set the KUBECONFIG environment variable:
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+```
+podman run -it --rm -e OC_LOGIN_SERVER=https://api.openshift.example.com:6443 \
+    -e OC_LOGIN_TOKEN=some-api-token \
+    quay.io/cloudservices/bonfire namespace list
+```
+
+It can also make use of a Kube config file by mounting it on the container's HOME:
+
+```
+podman run -it --rm -v ${KUBECONFIG:="$HOME/.kube/config"}:/opt/bonfire/.kube/config:Z,U \
```
bsquizz, line 95: Needs the `--userns` arg to ensure local kubeconfig permissions don't get altered after the container runs:

```suggestion
podman run -it --rm --userns=keep-id:uid=1000,gid=1000 -v ${KUBECONFIG:="$HOME/.kube/config"}:/opt/bonfire/.kube/config:Z,U \
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+    quay.io/cloudservices/bonfire namespace list
+```
+
+Although you can define the KUBECONFIG location using a variable:
+
+```
+podman run -it --rm -v $HOME/.kube/config:/opt/kubeconfig:Z,U \
```
bsquizz, line 102:
```suggestion
podman run -it --rm --userns=keep-id:uid=1000,gid=1000 -v $HOME/.kube/config:/opt/kubeconfig:Z,U \
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
 
 **Please note**: When using the Github token (classic), it **must have** the `read:project` scope for the template fetching to work!
 
+## Using Docker
```
bsquizz, line 70:
```suggestion
## Using a container image (podman/docker)
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+
+We provide a Docker image at quay.io/cloudservices/bonfire. The image wraps a Python virtual environment with **crc-bonfire** installed on it.
```
bsquizz, line 72:
```suggestion
We provide a container image at quay.io/cloudservices/bonfire. The image wraps a Python virtual environment with **crc-bonfire** installed on it.
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+
+In order to use Bonfire you have to provide either a valid KUBE_CONFIG or a URL and a TOKEN to connect to the target Openshift cluster.
```
bsquizz, line 74:
```suggestion
In order to use Bonfire you have to provide either a valid KUBECONFIG or a URL and a TOKEN to connect to the target OpenShift cluster.
```

README.md
```diff
@@ -65,6 +67,43 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
...
+
+The container's entrypoint already invokes the **bonfire** CLI with the arguments passed to the container, so you should be able to run any bonfire command you'd typically run by using:
```
bsquizz, line 76:
```suggestion
The container's entrypoint already invokes the **bonfire** CLI with the arguments passed to the container, so you should be able to run any bonfire command you'd typically run, for example:
```

PR #355, opened 2024-02-13
Count matching requester names as an optional valid ownership check

adamrdrew:
Today bonfire determines if a namespace is owned by the user requesting to re-use that namespace by comparing the namespace.name property to `oc whoami` - in firelink this check always fails. I want to be able to have a namespace ownership check pass during a namespace re-use request if the requester matches namespace.name. This patch represents the gentlest way I could see to do that. It shouldn't effect today's existing logic at all.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1136,7 +1136,7 @@ def _get_namespace(
 
     ns = None
     if requested_ns_name:
-        ns = _check_and_use_namespace(requested_ns_name, using_current)
+        ns = _check_and_use_namespace(requested_ns_name, using_current, requester)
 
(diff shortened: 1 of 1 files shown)


bsquizz (approved):
Works for now. We actually check if requester matches in `owned_by_me` but it is using an env var which isn't ideal for the firelink use case.

Also, had to double check that we'd never hit True on this or statement with `None == None` ... the `Namespace.requester` property can sometimes be `None` in certain situations -- but on the "making a reservation" side of things, it is never `None` due to us calling the `_get_requester` function at reservation time... `"bonfire"` is the default if nothing better can be determined.

PR #364, opened 2024-05-15
feat: import configmap from local directory at deploy time

akoserwal:
Adding a feature to import configmaps from a local directory at the deploy time

add flags: --configmaps-dir and --import-configmaps
same as secrets

```
bonfire deploy test --configmaps-dir /bonfire/testconfigmaps/ --import-configmaps true
```


bsquizz (approved):
LGTM, I wouldn't mind consolidating the code for secrets and configmaps into 1 set of functions (because they do nearly the same thing) but we can refactor later.

PR #381, opened 2024-06-21
(feat): Preserve trusted cpu/mem resource configurations

bsquizz:
Consider the CPU/memory configuration of a template as “trusted” if:

* The config value is defined using a template parameter.
* The parameter name follows the naming convention:
```
CPU_REQUEST_<NAME>
CPU_LIMIT_<NAME>
MEM_REQUEST_<NAME>
MEM_LIMIT_<NAME>
```

(#402 adds support for either `MEM_` or `MEMORY_`)

* The parameter's value is explicitly set on the deploy configurations for the component in the target environment.

Currently we only analyze this on ClowdApp/ClowdJob/ClowdJobInvocation -- in future once all teams make the necessary changes we'd probably want to circle back and turn this "type check" off so all objects are analyzed.

Example log output when finding trusted and untrusted configs:
```
2024-07-02 12:59:08 [   DEBUG] [          MainThread] value '${CPU_LIMIT_SERVICE}', regex r'\${(CPU_LIMIT[A-Z0-9_]+)}', matches=True, in params=True
2024-07-02 12:59:08 [   DEBUG] [          MainThread] value '${MEMORY_LIMIT_SERVICE}', regex r'\${(MEM_LIMIT[A-Z0-9_]+)}', matches=False, in params=False
2024-07-02 12:59:08 [   DEBUG] [          MainThread] deleted untrusted config at '.spec.deployments[2].podSpec.resources.limits.memory'
2024-07-02 12:59:08 [   DEBUG] [          MainThread] value '${CPU_REQUEST_SERVICE}', regex r'\${(CPU_REQUEST[A-Z0-9_]+)}', matches=True, in params=True
2024-07-02 12:59:08 [   DEBUG] [          MainThread] value '${MEMORY_REQUEST_SERVICE}', regex r'\${(MEM_REQUEST[A-Z0-9_]+)}', matches=False, in params=False
2024-07-02 12:59:08 [   DEBUG] [          MainThread] deleted untrusted config at '.spec.deployments[2].podSpec.resources.requests.memory'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,11 @@ Once you have audited your container CPU/mem usage in test environments and have
 If these steps are followed then your resource config is considered "trusted" and bonfire will not remove the requests/limits configurations.
 
 
+**NOTE:** a couple things that will change in the future:
+* The "host-inventory" app is trusted by default, but once their configurations are tweaked to be trusted we will remove this
+* Object types such as Deployment/StatefulSet/DaemonSet/etc. are not currently analyzed. In future once teams have audited all container resource configurations, we will begin to have bonfire check for trusted configurations on all object types.
+
+
 # Interactions with Ephemeral Namespace Operator
 
 `bonfire` creates/modifies `NamespaceReservation` resources on the cluster to reserve and release namespaces. The template for the object that bonfire applies into the cluster can be found [here](bonfire/resources/reservation-template.yaml)
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ As an example, typing `bonfire deploy host-inventory` leads to the host-inventor
 - [Examples for Common Use Cases](#examples-for-common-use-cases)
 - [Commonly Used CLI Options](#commonly-used-cli-options)
   - [Deploying/Processing](#deployingprocessing)
+  - [Trusted/Untrusted Resource Configurations](#trusteduntrusted-resource-configurations)
 - [Interactions with Ephemeral Namespace Operator](#interactions-with-ephemeral-namespace-operator)
 - [Interactions with Clowder Operator](#interactions-with-clowder-operator)
   - [ClowdEnvironments](#clowdenvironments)
@@ -120,7 +121,7 @@ Once logged in, you should be able to type `bonfire deploy advisor`
 
 This will cause `bonfire` to do the following:
 1. reserve a namespace
-2. fetch the templates for all components in the 'advisor' app and process them. By default, bonfire will look up the template and commit hash using the 'main'/'master' branch of each component defined in the app. It will look up the git commit hash for this branch and automatically set the `IMAGE_TAG` parameter passed to the templates. 
+2. fetch the templates for all components in the 'advisor' app and process them. By default, bonfire will look up the template and commit hash using the 'main'/'master' branch of each component defined in the app. It will look up the git commit hash for this branch and automatically set the `IMAGE_TAG` parameter passed to the templates.
 3. if ClowdApp resources are found, figure out which additional ClowdApps to fetch and process
 4. apply all the processed template resources into your reserved namespace
 5. wait for all the resources to reach a 'ready' state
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -210,6 +211,31 @@ bonfire deploy advisor --set-image-tag quay.io/cloudservices/advisor-backend=my_
...
+
+ The reason for this behavior is because many app templates tend to request far more CPU/memory in their template compared to what they actually need in a test environment.
+
+You can use the `--no-remove-resources` option to turn this behavior off for certain apps or components but the preferred approach is to audit your CPU/memory configurations and set up a resource configuration that bonfire trusts.
+
+Once you have audited your container CPU/mem usage in test environments and have determined efficient values that you want to use, you can indicate that bonfire should trust the CPU/Memory requests and limits by adhering to the following requirements:
+
+1. Any ClowdApp/ClowdJob/ClowdJobInvocation in your template asking for CPU/mem using requests or limits should use a parameter with a name formatted like this:
+* `CPU_REQUEST_<NAME>`
```
adamrdrew, line 225: I think if I was coming into this documentation blind I'd be a bit confused by this. I wouldn't know if `<NAME>` had to be some specific or consistent token, or if it could be arbitrary. Reading the code I know its any arbitrary combination of uppercase letters, numbers, and underscores, but reading this doc I wouldn't get that. Also, could be worth calling out what will pass the regex format - something like "where `<NAME>` is any text you want that is conforms to ... ". Finally, an example would be best. Some people will read it as written and do it right and get it, some will ask for clarification I think.

bonfire/processor.py
```diff
@@ -30,23 +30,82 @@ def _process_template(*args, **kwargs):
...
+    in_params = False
+    match = re.match(regex, value)
+    if match and match.groups()[0] in component_params:
+        in_params = True
+
+    log.debug(
+        "value '%s', regex r'%s', matches=%s, in params=%s", value, regex, bool(match), in_params
+    )
+
+    if match and in_params:
```
adamrdrew, line 52: Could you just do 

```
return match and in_params
```

I'm thinking that would solve the problem of no `False` return in this method and would remove the conditional.

bonfire/processor.py
```diff
@@ -30,23 +30,82 @@ def _process_template(*args, **kwargs):
...
+    Check for presence of a trusted param being used for the value.
+
+    A trusted parameter is defined as:
+    1. matching the expected regex pattern
+    2. the parameter value is set in the component's deploy config
+
+    Returns True if we determine this is a trusted value
+    """
+    in_params = False
+    match = re.match(regex, value)
```
adamrdrew, line 44: Do you need to maybe check that value and regex aren't `None`? I'm not sure. Just wondering in the absence of a type system we could end up with unexpected results of the params aren't what we expect.

bonfire/processor.py
```diff
@@ -30,23 +30,82 @@ def _process_template(*args, **kwargs):
...
+
+    Checks template to see if any resources of kind listed in 'config.TRUSTED_CHECK_KINDS'
+    are found. If so, analyzes the config on those resources to see if any need to be removed.
+    """
+    for obj in template.get("objects"):
+        kind = obj.get("kind")
+        if kind not in conf.TRUSTED_CHECK_KINDS:
+            continue
+
+        log.debug("checking resources on %s '%s'", kind, obj["metadata"]["name"])
```
adamrdrew, line 106: Should you use `gets` here to make sure `obj["metadata"]["name"]` exists? I could picture a scenario where we actually bomb out in the log message

bonfire/processor.py
```diff
@@ -30,23 +30,82 @@ def _process_template(*args, **kwargs):
...
+
+        A configuration is trusted if:
+        1. the value is defined using a template parameter
+        2. the parameter follows the proper syntax convention
+        3. the parameter is defined on the component in its deployment config
+
+    Checks template to see if any resources of kind listed in 'config.TRUSTED_CHECK_KINDS'
+    are found. If so, analyzes the config on those resources to see if any need to be removed.
+    """
+    for obj in template.get("objects"):
```
adamrdrew, line 101: maybe 

```
for obj in template.get("objects", []):
```

to be safe?

bonfire/processor.py
```diff
@@ -30,23 +30,82 @@ def _process_template(*args, **kwargs):
...
+
+def _remove_untrusted_configs(data, params, path=None, current_dict=None, current_key=None):
+    """
+    Locate configurations within 'data' and remove them if not trusted.
+
+    Checks to see if any config matching a path listed in 'config.TRUSTED_PARAM_REGEX_FOR_PATH' is
+    found within 'data' dictionary and ensures the regex matches and that the parameter value is
+    set on the component's deploy config.
+    """
+    if not path:
```
adamrdrew, line 64: You are setting `path=None` as the default argument value, could you just set `path=""` and avoid this conditional assign?

adamrdrew (dismissed):
LGMT - I pointed out some stuff, all very minor. You could merge it as is or make some changes if you like them. I would recommend the ones that deal with safety be incorporated but its all quite minor.

maknop (dismissed):
I can't say I found anything more than just the comments that @adamrdrew already left. Overall this looks like a really solid change. Greats tests, too!

bonfire/processor.py
bsquizz: Good point here. `regex` should never be `None` because we pull it out of the config, but I'll add a type check. `value` could possibly be null/empty so I'll check for that.

bsquizz:
Good feedback, thanks @adamrdrew

PR #427, opened 2024-10-01
feat: use local source for installing bonfire

Victoremepunto:


Dockerfile
```diff
@@ -1,24 +1,32 @@
...
-  -o oc.tar.gz && \
-  tar -C /usr/bin/ -xvzf oc.tar.gz oc kubectl && \
-  rm -f oc.tar.gz
+USER 1001
+WORKDIR $APP_ROOT
 
-USER bonfire
-WORKDIR /opt/bonfire
-ENV PATH="/opt/bonfire/.local/bin:$PATH"
+COPY --chown=1001:0 . .
```
bsquizz, line 24: This could potentially copy in the hidden `.` files right? Do you need to copy explicit paths here?

adamrdrew:
Other than what Brandon pointed out this LGTM

Dockerfile
Victoremepunto: yeah actually due to how we handle versions, I do - bonfire runs this scm dynamic versioning, I need the git repo metadata to get the right version 

I was thinking building the wheel and installing that instead - but there's not much of a difference anyways

Victoremepunto: that said, I still think a `.dockerignore` could really help keeping a bunch of not really needed files out, so I'll push that, thanks!

Victoremepunto: I added a multi-stage build to build a wheel and then copy and install that alone, please can you give it a second look?

bsquizz (approved):
Looks good!

PR #433, opened 2024-10-19
Add homebrew install instructions

alechenninger:
I created a homebrew tap because I am lazy and didn't want to mess with python virtual environments ;-).

The repo for the formula is here: https://github.com/redhatinsights/homebrew-bonfire

There are instructions on how to update it there, but it is easy enough to maintain so I'm happy to keep it up to date myself also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,13 @@ echo 'GITHUB_TOKEN=<your api token>' >> ~/.config/bonfire/env
 In addition, since some projects may be private, you will need to add the `repo` (full control of private repositories) scope too.
 for the template fetching to work!
 
+## Using homebrew
+
+```bash
+brew tap redhatinsights/bonfire https://github.com/RedHatInsights/homebrew-bonfire.git
+brew install bonfire
+```
+
 ## Using a container image (podman/docker)
 
 We provide a container image at quay.io/cloudservices/bonfire. The image wraps a Python virtual environment with **crc-bonfire** installed on it.


bsquizz (commented):
Thanks for adding! Tested it out, and it works, so I have no problem merging... I have some questions for possible enhancements. Would you prefer to merge this now and tweak things later or would it be easier for some of this to go in now? Let me know.

1. Should the formula just live in this repo or is it better to keep it in the separate one?
2. Could the formula update become part of a github release pipeline? (we have a github action that releases to pypi, could the formula be updated at that point in time?)
3. Is there a way to detect within the running app that bonfire was installed with homebrew? If so we might want to conditionally print a different message here: https://github.com/RedHatInsights/bonfire/blob/609172e7b23052c5e5bb9acadbfe86ff8e16902d/bonfire/utils.py#L505

alechenninger:
Thanks for reviewing!

> Should the formula just live in this repo or is it better to keep it in the separate one?

Looks like this is required:

> On GitHub, your repository must be named homebrew-something to use the one-argument form of brew tap. The prefix “homebrew-“ is not optional. (The two-argument form doesn’t have this limitation, but it forces you to give the full URL explicitly.)

https://docs.brew.sh/Taps

Though it does imply the instructions could be simplified per those docs.

> Could the formula update become part of a github release pipeline? (we have a github action that releases to pypi, could the formula be updated at that point in time?)

I think this is definitely possible! You would grab the latest source path, hash, and the run `brew update-python-resources Formula/bonfire.rb` and that should do the needful. That said I'm not sure I'll have time to tinker with that.

> Is there a way to detect within the running app that bonfire was installed with homebrew? If so we might want to conditionally print a different message here:

Good question. We could maybe touch a file to a known location during the install and check that?

bsquizz (approved):
I think it's good for now, enhancements can be made in future if we want :)

PR #438, opened 2024-11-14
feat: inject frontend context config map into pod spec

Hyperkid123:
The dynamic env name makes it difficult to properly mount FEO generated config map which holds the frontend context metadata to pods.

@psav and I had a discussion and today, this seems to be best for enabling new FEO features in ephemeral. We have briefly discussed some possible changes to FEO, Chrome backend, and clowder in the future, but right now we don't have many options.

Please let me know your thoughts.


florkbr (approved):
My 2 cents - I think this looks good as a stop-gap measure until we have a companion server to manage config.

bsquizz (approved):

PR #442, opened 2024-12-02
Merge remote and local app configs after setting git refs

### LINKED ISSUE #441: Ref always updated overriding local config value
I am struggling to make bonfire work with a local config file pointing to a remote repo as described in the [Readme](https://github.com/RedHatInsights/bonfire?tab=readme-ov-file#deploying-changes-changes-in-a-remote-git-branch).

To me is seems that the provided `ref` value gets always overriden by the ref value from the `--ref-env` or the fallback one at https://github.com/RedHatInsights/bonfire/blob/master/bonfire/qontract.py#L382

As a user, I would like to be able to use whatever reference env I set via `--ref-env` for most of my components except for the ones overridden by my local config.

When using the local config to point to a local host things work as expected, I guess because the `ref` is ignored.

bsquizz:
Fixes #441

If a git ref is set on a component in the local config, it should take precedence over the remote config.

* Wait to merge remote config with local config until AFTER we set the git refs for all components
* Add better debug logging around the config merge
* Tweak unit tests to properly test this scenario

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -951,6 +951,14 @@ def _get_apps_config(
         log.info("fetching apps config using source: %s", source)
         apps_config = get_appsfile_apps(config)
 
+    # handle git ref/image substitutions if reference environment was provided
+    if ref_env:
+        apps_config = sub_refs(apps_config, ref_env, fallback_ref_env, preferred_params)
+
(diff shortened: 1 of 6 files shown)


adamrdrew (approved):
This looks good to me. I have no changes to suggest.

PR #456, opened 2025-02-03
Frontend context name ephemeral value

Hyperkid123:
This change is required for the Frontend operator features and is a pre-requisite for opening the migrations to the new features for HCC tenants.

### Why
By default, FEO generates a config map to the Frontend namespace(s). In stage/prod, this config map is also injected into other namespaces(s) with a deterministic name and used there. However because in EE all deployments are in the same namespace, we need to set the parameter `FRONTEND_CONTEXT_NAME` set to the `clowd_env` value in order to properly inject the config map value into specific pods (the deployment use the FRONTEND_CONTEXT_NAME env variable to get the name of the config map).

There is already one place in the processor.py where this parameter is configured however, it turns out it's needed in one additional place.

We will need a new version of bonfire to be released so it can be added to the Frontend migration guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,19 @@ podman run -it --rm --userns=keep-id:uid=1000,gid=1000 -v $HOME/.kube/config:/op
     quay.io/cloudservices/bonfire namespace list
 ```
 
+# Local development
+
+To install local changes and use local version of bonfire, switch to the root directory of this repository, switch to virtual environment and install bonfire package with local changes:
+
+```bash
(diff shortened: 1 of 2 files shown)


Hyperkid123:
Looking at the Konflux failures, we will most likely have to merge: https://github.com/RedHatInsights/bonfire/pull/452

Hyperkid123:
Additional manual Konflux updates are required: https://github.com/RedHatInsights/bonfire/pull/457

gburges (approved):

PR #465, opened 2025-03-14
fix konflux tasks

maknop:
**Description**

The ordering of a couple tasks need to be fixed in order for the `on-pull` and `on-push` pipelines to pass.


adamrdrew (approved):

adamrdrew (approved):
🤷

PR #475, opened 2025-07-09
Pass NAMESPACE param if unset, fix bugs in resource waiter

bsquizz:
This PR handles 3 things:

* If NAMESPACE parameter is unset, pass it into templates during processing (similar to how we handle IMAGE_TAG)
* Fix bugs found with resource waiter:
    * If there are no ClowdApps/ClowdEnvironments, the ResourceWatcher did not have enough time to fetch resources from the kubernetes API. This caused `_all_resources_ready` to think that there were no additional resources left to wait on. This fix is to fetch all resources initially with a "blocking" call before starting the background thread.
    * `wait_for_all_resources` does not need to enclose `_all_resources_ready` in a `wait_for()` function. This is because `_all_resources_ready` was modified some time ago to keep lowering the timeout as it moves through each "phase". By using `wait_for` in `_all_resources_ready`, if the `_all_resources_ready` returned False before `timeout` was hit (even by a small fraction), `_all_resources_ready` would get re-executed again (and wait for another 10 minutes). That's not the behavior we want. 
* Upgrade ocviapy to v1.4.0 (this adds a new function which helps with the above)

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1023,6 +1023,7 @@ def _process(
     local,
     frontends,
     preferred_params,
+    namespace,
 ):
     apps_config = _get_apps_config(
         source,
@@ -1051,6 +1052,7 @@ def _process(
         component_filter,
         local,
         frontends,
+        namespace,
     )
     return processor.process()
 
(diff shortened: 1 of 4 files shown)


bsquizz:
The changes for resource waiting will need some testing:
1. Test to ensure a broken ClowdEnvironment resource times out as expected
2. Test to ensure a broken ClowdApp resource times out as expected
3. Test the "happy path" for a namespace that has ClowdApps

PR #476, opened 2025-07-15
Support immediate exit upon hitting a status error, add `--defer-status-errors` CLI option

bsquizz:
As of ocviapy v1.5.0, there is now support for a StatusError to be raised whenever an error such as ImagePullBackOff is encountered. This allows bonfire to exit immediately if an "unrecoverable" error is found during the "wait on namespace resources to be ready" logic (instead of wasting time waiting for the full timeout period).

You can still invoke bonfire with `--defer-status-errors` set if you want to use the old behavior. This will ignore "status errors" such as an ImagePullBackOff until the very end ... instead, it will wait for the resources in the namespace to be ready. If a timeout is hit, we will print the "bad" status conditions at that time.


bonfire/bonfire.py
```diff
@@ -1349,19 +1379,6 @@ def _cmd_config_deploy(
 
     clowd_env = _get_env_name(ns, clowd_env)
 
-    def _err_handler(err):
```
bsquizz, line 1352: This func has now become `_deploy_err_handler` above.

adamrdrew:
This looks good. I see no problems and ARTIFICIAL INTELLIGENCE concurs. LGTM

PR #482, opened 2025-07-21
add cli parameter that exclude components to deploy

JuanmaBM:
This PR adds a new CLI argument `--exclude-components`, which accepts a comma-separated list of components to be skipped by the processor so they are not deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,11 @@ bonfire deploy advisor --set-parameter advisor-backend/REPLICAS=3
 bonfire deploy advisor --set-image-tag quay.io/cloudservices/advisor-backend=my_tag
 ```
 
+* Exclude components of a app
+```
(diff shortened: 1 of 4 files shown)


bsquizz (changes requested):
I spotted a problem while testing this. I think ideally what we want `--exclude-components` to do is put all of the configurations into a namespace EXCEPT for the component you've asked to exclude

Right now, if we use `--exclude advisor-backend` it will cause us to not pick up any of the dependencies defined on the advisor-backend ClowdApp and we actually end up with a completely empty namespace.

I think what will work is... if the component_name is in `exclude_components`, we will NOT add the processed component to the "k8s_items" list here: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/processor.py#L832

This would omit the template configs from appearing in the final output, but still call `_handle_dependencies` for this component.

PR #486, opened 2025-07-30
Make sure the 'org' value uses %2F for / when formatted into paths

PaulWay:
For no readily apparent reason, we recently had to create a project with the
'org' part of the path as 'insights-platform/insights-operations'.  This unfortunately
causes gitlab to get confused, because it parses that as the 'insights-platform'
org and then the rest of the line doesn't match what it expects.  So we need to
URL escape the `/` in the 'org' value.

### DIFF
--- a/bonfire/qontract.py
+++ b/bonfire/qontract.py
@@ -238,9 +238,15 @@ def _add_component(
     host = "github" if "github" in url else "gitlab"
 
     try:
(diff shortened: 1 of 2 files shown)


bsquizz:
After #487, I'm able to `bonfire process advisor` again ... so did that PR take care of this problem?

PR #492, opened 2025-08-04
adds uv tool install instructions

iamkirkbater:
I found this installation method to be extremely simple, so I wanted to add it to the documentation. This is for the casual end-user who may not want to muck around with virtual environments† and python versions and the dependencies between those.

† - "may not want to muck around with virtual environments" meaning at face value. UV does this under the hood, but that's abstracted to the end-user and this should "just work" just like any other system binary.


bsquizz (approved):
Looks good, thanks!

Issue #217, opened 2022-05-13
RFE: support namespace as part of root bonfire command


Victoremepunto:
I hope this makes sense!

 I'm very used to provide a namespace to `oc` as part of the first options of every command, so if I have to issue multiple commands, I can repeat last command keeping the namespace variable part. 

I cannot do this on `bonfire` because apparently `namespace` is part of specific commands or sub-commands like `deploy` or `namespace extend` and not a top level object.

While I understand `oc` is a general purpose tool for Openshift, specifically, I think it would improve the UX with Bonfire. However it's true I don't know if there are enough Bonfire operations specific to a namespace in particular for this to make sense, but ... if it does, here's an example of usage to illustrate what I mean:

This today doesn't work:
```sh
bonfire -n $EPHEMERAL_NAMESPACE deploy rbac
Usage: bonfire [OPTIONS] COMMAND [ARGS]...
Try 'bonfire -h' for help.

Error: no such option: -n
```

You should be doing this instead:
```sh
bonfire deploy -n $EPHEMERAL_NAMESPACE rbac
```

Now if I want to  extend the duration of one of them I have to write it like this

```sh
bonfire namespace extend -d 1h $EPHEMEREAL_NAMESPACE
```

Whereas both commands could be like:
```sh
bonfire -n  $EPHEMEREAL_NAMESPACE deploy foobar
bonfire -n  $EPHEMEREAL_NAMESPACE namespace extend -d 1h
```

bsquizz:
This is also a quirk with the click CLI library I haven't spend time trying to solve yet: I'd like to be able to provide `--debug` at any point in the CLI but it only works if applied at the root level, e.g:

`bonfire --debug deploy <args>` works but `bonfire deploy <args> --debug` does not

I think it is possible to pass args between the command groups by sharing context but I also am not 100% sure how to do it best yet.

PR #498, opened 2025-08-07
Namespace option at Root or subcommands

mjholder:
The purpose of this PR is to address https://github.com/RedHatInsights/bonfire/issues/217.
To do this we add the namespace option to the root command and remove the required status from the sub commands. For those sub commands we do a check if the namespace was provided at the root or the subcommand.

Unfortunately we can't do the same for the debug option. In click the root command gets processed first, you can't pass the context back up to the root.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -116,17 +116,16 @@ def click_exception_wrapper(command):
 
 
 _global_options = [
-    click.option("--debug", "-d", help="Enable debug logging", is_flag=True, default=False),
     click.option("--namespace", "-n", help="Namespace to use", default=None),
 ]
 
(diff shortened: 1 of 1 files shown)


bsquizz (changes requested):
This looks good. Claude pointed this out which I agree with: _The pattern `_namespace = namespace or ctx.obj.get("namespace")` is repeated multiple times. Consider extracting this into a helper function._

I'd suggest something like this:

  ```python
  def get_namespace_from_context(ctx, namespace, required=False):
      """Get namespace from local option or global context."""
      _namespace = namespace or ctx.obj.get("namespace")
      
      if not _namespace and required:
          raise click.UsageError("namespace is required (use --namespace/-n)")
       
      return _namespace
  ```

PR #500, opened 2025-08-07
Allow fine-grained control over dependency removal

awood:
In some cases, uses may want to be selective about dependency removal. For example, if some dependencies are being mocked during application testing (e.g. with WireMock), the user might still wish to retain other dependencies.

This patch changes the `--no-remove-dependency` and `--remove-dependency` options to accept a slash after a component name followed by a dependency.  If no slash follows, the component name, all dependencies will be operated on.

Obviously this creates a resolution problem if the user provides something like` --remove-dependencies foo --no-remove-dependencies foo/bar`.  The code follows two rules: "specific beats general" and "keep by default".  So in the example, the result would be the removal of all dependencies except for foo/bar.  The resolution is handled through set operations.  Or in the example `--remove-dependencies component/b --keep-dependencies component/c` with an application that had the dependencies a, b, and c, the result would be the retention of the dependencies a and c.

This patch also adds substantial validation to catch cases that may be technically resolvable but are more ambiguous.  For example, `--remove-dependencies foo --remove-dependencies foo/bar`.  While this case is resolvable without too much thought, I didn't want to introduce additional complexity into what is already a fairly complex process.

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -366,27 +366,39 @@ def _validate_split_equals(ctx, param, value):
         raise click.BadParameter(msg)
 
 
-def _translate_to_obj(value_list):
+def _translate_to_obj(value_list, param_name):
     apps = []
     components = []
     select_all = False
+    # By default, we don't want to remove any dependencies
+    selector_options = ["remove_dependencies", "no_remove_dependencies"]
     for value in value_list:
         if value == "all":
             # all is a special keyword
             select_all = True
         elif value.startswith("app:"):
             apps.append(value.split(":")[1])
+        elif "/" in value and param_name not in selector_options:
+            raise click.BadParameter("dependency specifiers are not allowed for this option")
         else:
             components.append(value)
-    return AppOrComponentSelector(select_all, apps, components)
+    try:
+        return AppOrComponentSelector(
+            select_all,
+            apps,
+            components,
+        )
+    except ValueError as e:
+        raise click.BadParameter(str(e))
 
 
 def _app_or_component_selector(ctx, param, this_value):
     if any([val.startswith("-") for val in this_value]):
         raise click.BadParameter("requires a component name or keyword 'all'")
 
-    # check if 'app:' syntax has been used and translate input values to apps/components dictionary
-    this_value = _translate_to_obj(this_value)
+    # check if 'app:' syntax or "/dependency" syntax has been used and
+    # translate input values to apps/components dictionary
+    this_value = _translate_to_obj(this_value, param.name)
 
     opposite_option = {
         "remove_resources": "no_remove_resources",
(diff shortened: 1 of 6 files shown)


awood:
It doesn't like I can put this in draft, but hold on this for a little bit.  There's a regression where components are being included when they shouldn't be.

awood:
> It doesn't like I can put this in draft, but hold on this for a little bit. There's a regression where components are being included when they shouldn't be.

I believe this is working correctly.  I was a little confused because the version of Bonfire I was using as a control in my experiments was an old one.  It looks like Bonfire's behavior around dependencies has changed recently to include dependencies from components even if the component is not on the component list.  Looking at the differences in the JSON output from process:

```
$ bonfire process rhsm --source=appsre --ref-env insights-stage --component swatch-contracts --remove-dependencies swatch-contracts --no-remove-dependencies swatch-contracts/swatch-database --component swatch-database --component rbac > branch-fine-remove-deps.json
$ jq -c '.items.[] | select(.kind | contains("ClowdApp")) | { name: .metadata.name, deps: .spec.dependencies } ' < branch-fine-remove-deps.json | sort

# and on the master branch

$ bonfire process rhsm --source=appsre --ref-env insights-stage --component swatch-contracts --remove-dependencies swatch-contracts --component swatch-database --component rbac > master-remove-deps.json
$ jq -c '.items.[] | select(.kind | contains("ClowdApp")) | { name: .metadata.name, deps: .spec.dependencies } ' < master-remove-deps.json | sort
```

Has a difference of

```diff
8c8
< {"name":"swatch-contracts","deps":["swatch-database"]}
---
> {"name":"swatch-contracts","deps":null}
```

If we use the master branch and run

```
$ bonfire process rhsm --source=appsre --ref-env insights-stage --component swatch-contracts --component swatch-database --component rbac > master-keep-deps.json
$ jq -c '.items.[] | select(.kind | contains("ClowdApp")) | { name: .metadata.name, deps: .spec.dependencies } ' < master-keep-deps.json | sort
```

And do a diff with the branch `jq` results, the result is

```diff
8c8
< {"name":"swatch-contracts","deps":["swatch-database"]}
---
> {"name":"swatch-contracts","deps":["swatch-database","swatch-tally","rbac","export-service"]}
```

So the code is correctly limiting the dependencies.  I actually want to be able to be a little stricter with Bonfire not pulling in extra deps from unlisted components, but that's a PR for another day.

Now, there is one slight alteration, and that's on this branch if you use `--remove-dependencies swatch-contracts`, the end result of that `jq` query will be `{"name":"swatch-contracts","deps":[]}`.  I don't believe that should be an issue, but I would like confirmation that an empty `dependencies` block in the YAML will be equivalent to the block not being present at all.  The empty list is there because, the code now goes in after the fact and deletes items from the `dependencies` section and if a broad removal is requested, it just removes everything from the list but doesn't set the entry to null like before.

PR #503, opened 2025-08-15
Support for specifying team name upon reserving a namespace

maknop:
# Purpose

The purpose of this change is to prompt the user for their team name which will get added on to the reserved ephemeral namespace. The new field will be added to `spec.team` on the `NamespaceReservation` resource. 

This change is intended to go along with [Ephemeral Namespace Operator MR #387](https://github.com/RedHatInsights/ephemeral-namespace-operator/pull/387)

Assisted-By: Cursor v1.2.4 / claude-4-sonnet

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1243,9 +1243,7 @@ def _check_and_reserve_namespace(name, requester, duration, pool, timeout, local
 
     team = ""
     if pool == "ai-development":
-        log.info("------------------------------- This pool exists! -------------------------------------")
-        team = input("What team are you on? ")
-        log.info(f"user is on team: {team}")
+        team = input("You are about to reserve a namespace from the 'ai-development' pool; What team are you on? ")
 
     if pool not in get_namespace_pools():
         _error(f"namespace pool '{pool}' does not exist on this cluster")
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1243,7 +1243,9 @@ def _check_and_reserve_namespace(name, requester, duration, pool, timeout, local
 
     team = ""
     if pool in conf.AI_SPECIFIC_POOLS:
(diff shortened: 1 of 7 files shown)


bonfire/bonfire.py
```diff
@@ -237,6 +237,12 @@ def _validate_reservation_duration(ctx, param, value):
         default=None,
         help="Name of the user requesting a reservation",
     ),
+    click.option(
+        "--team",
+        type=str,
+        default="",
+        help="Team name associated with requester for",
```
bsquizz, line 244:
```suggestion
        help="Team name associated with reservation requester",
```

bonfire/processor.py
```diff
@@ -403,6 +404,7 @@ def process_reservation(name, requester, duration, pool=None, template_path=None
             requester = "bonfire"
 
     params["REQUESTER"] = requester
+    params["TEAM"] = team
```
bsquizz, line 407:
```suggestion
    params["TEAM"] = team or ""
```

bsquizz (changes requested):

bonfire/config.py
```diff
@@ -22,6 +22,7 @@
 DEFAULT_CONFIGMAPS_DIR = get_config_path().joinpath("configmaps")
 
 DEFAULT_NAMESPACE_POOL = "default"
+AI_SPECIFIC_POOLS = ["ai-development"]
```
bsquizz, line 25: Are we doing anything with this var anymore?

bsquizz: I think previously you were using this to check if the user is trying to make a reservation in the 'ai-development' pool but they have not specified a team. I think that is a good check to keep around. In `_check_and_reserve_namespace`, you could check if the requested pool is in `conf.AI_SPECIFIC_POOLS` (perhaps rename this to `POOLS_REQUIRING_TEAM`) and if team is not defined, throw an error.

bsquizz (dismissed):

bsquizz (approved):

PR #507, opened 2025-08-26
Track component apply

bsquizz:
The following scenario was not working:

* app 'myapp' has components: 1, 2, 3
* component 1 depends on component 2.
* component 3 also depends on component 2.
* bonfire is run with `bonfire process myapp --component 3`

Component 2 does not initially get applied because the 'component_skip_check' removes it (as it should)

When it comes time for component 2 to get pulled by component 3, bonfire thinks it has already processed component 2 and skips it.


bsquizz:
#500 will address this fix

PR #508, opened 2025-08-29
Fix dependency processing in components with multiple ClowdApps

awood:
The code was erroneously limiting dependency processing to ClowdApps that match the component name exactly.  While this is often the case, it is possible for a component to contain multiple ClowdApps.


bsquizz (approved):

bsquizz:
v6.6.0 had this change. At this point v6.7.2 is the latest, so you should definitely be seeing this fix if you are up-to-date

PR #512, opened 2025-10-06
Do not process components that are going to be skipped

ikerreyes:


awood:
Try

```patch
diff --git i/bonfire/processor.py w/bonfire/processor.py
index 5273e96..01bff7e 100644
--- i/bonfire/processor.py
+++ w/bonfire/processor.py
@@ -990,6 +990,12 @@ class TemplateProcessor:
             # will cause the code to go into self._handle_dependencies later on.
         else:
             log.info("--> processing component %s", component_name)
+            skip_further_processing = self._component_skip_check(component_name, dependency_chain)
+            if skip_further_processing:
+                log.info("skipping component '%s', %s", component_name, skip_further_processing)
+                return
+            else:
+                should_apply = True
             items = self._get_component_items(component_name)
 
             # ignore frontends if we're not supposed to deploy them
@@ -1001,6 +1007,7 @@ class TemplateProcessor:
                 items = []
 
             processed_component = ProcessedComponent(component_name, items)
+            processed_component.should_apply = should_apply
             self.processed_components[component_name] = processed_component
             skip_further_processing = self._component_skip_check(component_name, dependency_chain)
             if skip_further_processing:
```

It's the call to `_get_component_items` that was causing your failure.

bsquizz:
Handled via #513

PR #549, opened 2026-04-10
Add MCP server and shared library for ephemeral environments

gurnben:
### What

This PR adds two new packages to the bonfire repository:

- **`bonfire_lib`** — a shared Python library that extracts bonfire's reservation lifecycle logic (reserve, release, extend, list, status, describe) and rewrites it against the `kubernetes` Python client directly, eliminating the `oc` binary dependency for these operations.
- **`bonfire_mcp`** — an [MCP server](https://modelcontextprotocol.io/) that exposes 8 ephemeral environment tools to any MCP-compatible AI agent (Claude, GPT, Cursor, etc.) via stdio transport.

It also rewires the bonfire CLI's `namespace reserve`, `namespace release`, `namespace extend`, and `namespace describe` commands to delegate to `bonfire_lib` instead of using `ocviapy` directly.

### Why

AI coding agents increasingly need live infrastructure — deploying apps, running integration tests, reproducing production topologies. Today, ephemeral environments are only accessible via the bonfire CLI or raw CRD manipulation. There's no standard interface for agents to programmatically manage ephemeral resources.

The [design doc](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/engineering-productivity/design/design-ephemeral-mcp-server.md) and [ADR-002](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/engineering-productivity/decisions/adr-002-ephemeral-mcp-server.md) describe the full rationale. The key architectural decisions:

1. **Shared library extraction** — bonfire's core reservation logic was already well-separated from the CLI layer. Extracting it into `bonfire_lib` lets both the CLI and MCP server consume the same proven logic without duplication.
2. **`kubernetes` Python client instead of `ocviapy`** — the shared library and MCP server have zero dependency on the `oc` binary. This makes them portable to containers, CI pipelines, and dev workstations without OpenShift tooling. The bonfire CLI retains `ocviapy` for app/component template processing (`oc process`), which is out of scope.
3. **Jinja2 templates replacing `oc process`** — the 3 local core resource templates (`NamespaceReservation`, `ClowdEnvironment`, `ClowdJobInvocation`) are now Jinja2 YAML files rendered in Python, removing the last `oc` dependency from the reservation lifecycle path.

### What's included

| Area | Details |
|------|---------|
| `bonfire_lib/` | `config.py` (Settings dataclass), `k8s_client.py` (EphemeralK8sClient with 3 auth modes), `core_resources.py` + `templates/` (4 Jinja2 templates), `reservations.py`, `pools.py`, `status.py`, `clusters.py`, `resource_types.py`, `utils.py` |
| `bonfire_mcp/` | `server.py` (8 MCP tools with polymorphic namespace/cluster dispatch), `auth.py` (preflight checks mirroring `containers/kubernetes-mcp-server`), `formatters.py`, `__main__.py`, `README.md` |
| CLI rewiring | `bonfire/namespaces.py` delegates reserve/release/extend/describe to `bonfire_lib` |
| Package separation | Base deps are only `kubernetes`, `jinja2`, `pyyaml`. CLI deps (`ocviapy`, `click`, `sh`, etc.) moved to `[cli]` extra. `pip install crc-bonfire[mcp]` does not pull `ocviapy`. |
| Tests | 326 passing (113 bonfire_lib + 59 bonfire_mcp + 154 existing CLI), 6 integration tests skipped without cluster access |

### MCP tools

| Tool | Type | Description |
|------|------|-------------|
| `ephemeral_list_pools` | Read | List namespace/cluster pools with capacity stats |
| `ephemeral_reserve` | Mutate | Reserve a namespace (sync) or cluster (async) |
| `ephemeral_status` | Read | Get reservation status by name or namespace |
| `ephemeral_extend` | Mutate | Add duration to an active reservation |
| `ephemeral_release` | Mutate | Release a reservation |
| `ephemeral_list_reservations` | Read | List reservations, filterable by requester |
| `ephemeral_describe` | Read | Namespace details (ClowdApps, frontends, keycloak creds) |
| `ephemeral_get_kubeconfig` | Read | Fetch kubeconfig for a ROSA HCP cluster (Phase 3) |

### ROSA HCP cluster support (Phase 3 — client-side only)

The Python client-side code for cluster reservations (`clusters.py`, `resource_types.py`, `clusterreservation.yaml.j2`, cluster MCP tool dispatch) is implemented and mock-tested. The ENO CRDs and controllers (`ClusterReservation`, `ClusterPool`) are not yet built — that's Go work in `ephemeral-namespace-operator` blocked on ADR-001 finalization. The cluster tools gracefully degrade (404/403 on missing CRDs is handled as a warning, not a crash).

### Breaking changes

- `kubernetes>=29.0.0` and `jinja2>=3.1.0` are now base dependencies of `crc-bonfire` (previously not required).
- CLI-only dependencies (`ocviapy`, `click`, `sh`, etc.) are now under the `[cli]` optional extra. Users who install with `pip install crc-bonfire` will no longer get these automatically — use `pip install crc-bonfire[cli]` for the full CLI experience, or `pip install crc-bonfire[cli,mcp]` for everything.

### Not included

- Streamable HTTP transport (stdio only for now)
- App/component deployment via MCP (remains CLI-only, depends on `oc process`)
- ENO-side ROSA HCP controllers (separate repo, separate PR)
- `release_all` / session cleanup tools

### How to test

```bash
pip install -e ".[cli,test,mcp]"
pytest tests/ -q                            # 326 passed, 6 deselected
pytest -m integration                       # requires KUBECONFIG with ephemeral CRD access
bonfire-mcp                                 # starts MCP server on stdio
```

💘 Generated with Crush


gurnben:
Tests passed locally!

gurnben:
## Code Review Fixes (bb1afd4)

Hi — I'm an AI assistant (Claude via Crush) that reviewed this PR against the [design document](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/engineering-productivity/design/design-ephemeral-mcp-server.md) and wrote the following fixes. All 328 unit tests and 6 integration tests pass against the live `crc-eph` cluster.

### Bug fix

- **MCP error signaling** — Error handlers were setting `isError=True` on `TextContent`, but the MCP SDK ignores that field there. Errors are now returned as `CallToolResult(isError=True)`, which the SDK propagates correctly to agents. Affected all 5 exception handlers in `server.py`.

### Packaging / config

- **`requires-python`** — Changed `>=3.6` → `>=3.10`. The codebase uses `str | None` union syntax (PEP 604) throughout `bonfire_lib/`, which requires 3.10+. The old floor would produce `SyntaxError` on 3.6–3.9.
- **Duplicate pytest config** — Deleted `pytest.ini` and consolidated `log_cli`, `log_level`, `asyncio_mode`, `markers`, and `addopts` into `pyproject.toml [tool.pytest.ini_options]`. Eliminates the `WARNING: ignoring pytest config in pyproject.toml!` on every test run.
- **`asyncio_mode = "strict"`** — Pinned explicitly so future `pytest-asyncio` version changes don't break async test collection.

### Architecture cleanup

- **Removed dead `ResourceType` enum** — `bonfire_lib/resource_types.py` was defined per the design but never imported anywhere. Deleted until needed.
- **Moved cluster reservation list/format to proper modules** — `_list_cluster_reservations()` moved from `server.py` → `clusters.list_cluster_reservations()`. `_format_cluster_reservation_list()` moved from `server.py` → `formatters.format_cluster_reservation_list()`. This restores the clean separation where `server.py` is pure dispatch.
- **Deduplicated polling logic** — `reservations.py` had a private `_wait_for_namespace()` that reimplemented the same polling as `status.wait_on_reservation()`. Now delegates to the single implementation in `status.py`.

### Documentation

- **MCP README** — Rewrote the tools table and parameters section to document all 8 tools (was missing `ephemeral_get_kubeconfig` and the `type` parameter on 5 tools). Added a cluster workflow example alongside the existing namespace example.

### Minor fixes

- **`datetime.utcnow()` deprecation** — `bonfire/namespaces.py` now uses `datetime.now(datetime.UTC)` (eliminates pytest deprecation warning).
- **`__all__` exports** — Added to `bonfire_lib/__init__.py` to make the public API explicit.
- **Bare `except` in test fixture** — Integration test cleanup fixture now catches `Exception as exc` and logs a warning instead of silently swallowing errors.

pyproject.toml
```diff
@@ -24,6 +24,27 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
+    "PyYAML",
+    "kubernetes>=29.0.0",
+    "jinja2>=3.1.0",
+]
+
+[project.scripts]
```
JuanmaBM, line 32: The package metadata was changed so that core runtime dependencies are only PyYAML, kubernetes, and jinja2, while the Bonfire CLI stack (click, ocviapy, gql, etc.) lives under [project.optional-dependencies] cli.

The bonfire console script still targets bonfire.bonfire:main_with_handler, which imports the full CLI dependency graph at import time. As a result, pip install crc-bonfire without extras may fail when users run bonfire, or they may not get a working CLI.

The root README still documents:
```
pip install crc-bonfire
```

which no longer matches the intended install path if CLI users must use extras (e.g. pip install "crc-bonfire[cli]").

Expected outcome

- Either restore backward-compatible defaults (CLI deps in base dependencies, or another agreed pattern), or
- Update all install documentation (README, PyPI description, internal docs) and release notes so users explicitly install crc-bonfire[cli] (and [mcp] / [test] where relevant), with no silent breakage for existing workflows.

bonfire_mcp/server.py
JuanmaBM, line 1: In bonfire_mcp/server.py (line 434), the call_tool handler returns CallToolResult with isError=True (via _error_result) for validation failures, FatalError, timeouts, etc.

For an unknown tool name, the else branch returns plain TextContent only, without isError=True:
```
return [TextContent(type="text", text=f"Unknown tool: {name}")]
```

That is inconsistent with other error paths and may cause MCP clients to treat the response as success.

Expected outcome:
- Return the same shape as other errors, e.g. _error_result(f"Unknown tool: {name}"), so unknown tools are reported with isError=True and clients can handle them uniformly.

bonfire_lib/status.py
```diff
@@ -0,0 +1,216 @@
...
+    try:
+        cm = client.get_configmap("console-public", "openshift-config-managed")
+        if cm:
+            return cm.get("data", {}).get("consoleURL")
+    except Exception as err:
+        log.debug("unable to obtain console url: %s: %s", err.__class__.__name__, err)
+    return None
+
+
+def describe_namespace(client: EphemeralK8sClient, namespace: str) -> dict:
```
JuanmaBM, line 149: Describe was moved from bonfire/namespaces.py (via ocviapy / get_json) to bonfire_lib.status.describe_namespace (Kubernetes dynamic client + CoreV1Api).

For the happy path, the flow is unchanged: load namespace → verify operator label → count ClowdApps / Frontends → read FrontendEnvironment → read Keycloak secret → build CLI/JSON output.

The Problem:
Two implementation differences can change behavior when something goes wrong, compared to the previous describe:

1. Resource resolution: The new client resolves CRs through a fixed API version (cloud.redhat.com/v1alpha1 in EphemeralK8sClient). The old path relied on oc / get_json, which uses server discovery. If a cluster ever exposes these kinds under a different version or group, behavior may diverge from the legacy CLI.
2. Error visibility on ClowdApp / Frontend list: ClowdApp and Frontend listing is wrapped in broad try / except Exception handlers that fall back to empty lists. An API failure can therefore surface as “0 deployed” instead of propagating an error, whereas describe_namespace on master did not wrap those get_json calls the same way—failures were more likely to bubble up (depending on get_json behavior).

pyproject.toml
gurnben: Fixed in 7df8c61. Updated both install instructions in the README:
- `pip install crc-bonfire` → `pip install crc-bonfire[cli]`
- `pip install -e .` → `pip install -e ".[cli]"`

The Dockerfile and CI workflow were already updated in earlier commits (`d82b084` and `bda7b8b` respectively).

bonfire_mcp/server.py
gurnben: Good catch. Fixed in 7df8c61 — the `else` branch now returns `_error_result(f"Unknown tool: {name}")` for consistent `isError=True` signaling across all error paths. Test updated to assert `result.isError is True`.

bonfire_lib/status.py
gurnben: Both points are valid. Fixed in 7df8c61 — added `log.warning` to all three exception handlers in `describe_namespace` (ClowdApp listing, Frontend listing, FrontendEnvironment lookup) so API failures are visible in debug output rather than silently returning empty/default values.

The broad `except` is kept intentionally: these CRDs are optional (not every cluster has the Frontend CRD installed, for example), so crashing the entire describe output on a missing CRD would be worse than showing "0 deployed" with a warning in the logs. The fixed API version (`cloud.redhat.com/v1alpha1`) matches what ENO registers — if a cluster exposes these under a different group, the warning will surface the 404 clearly.

mjholder (approved):
LGTM! I tested it locally and don't see any regression. I also tested out a couple parts of the MCP and it worked without an issue

PR #572, opened 2026-05-14
Make cli non-optional

bsquizz:
Right now if you want the cli, you need to run `pip install crc-bonfire[cli]`, which is an introduced bug.


mjholder (approved):
LGTM

PR #576, opened 2026-05-26
fix: wrong k8s client configuration

JuanmaBM:
Fix a problem with the k8s configuration

### DIFF
--- a/bonfire_lib/k8s_client.py
+++ b/bonfire_lib/k8s_client.py
@@ -81,12 +81,15 @@ class EphemeralK8sClient:
             # This ensures that explicit `oc login` credentials are used even when
             # running inside a pod (e.g., GitLab CI runners).
             try:
+                # Use explicit configuration instead of relying on global state
+                kube_config = client.Configuration()
                 config.load_kube_config(
                     config_file=kubeconfig_path,
                     context=context,
+                    client_configuration=kube_config,
                 )
                 # Verify that the kubeconfig actually works by probing /apis
-                test_client = client.ApiClient()
+                test_client = client.ApiClient(kube_config)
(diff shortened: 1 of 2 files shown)


bonfire_lib/k8s_client.py
```diff
@@ -81,12 +81,15 @@ def __init__(
             # This ensures that explicit `oc login` credentials are used even when
             # running inside a pod (e.g., GitLab CI runners).
             try:
+                # Use explicit configuration instead of relying on global state
```
mshriver, line 84: Looks like this is now following the explicit passing demonstrated here:
https://github.com/kubernetes-client/python/blob/master/kubernetes/README.md#getting-started

Though I don't quite see in the Configuration class where defaults are applied from stored kubeconfig.

JuanmaBM: Configuration() creates an empty/default object. load_kube_config()                                                                                                                                                                  
  then mutates it by reading the kubeconfig file and populating host, api_key,                                                                                                                                                                
  ssl_ca_cert, etc. Same with load_incluster_config(), it reads from                                                                                                                                                                         
  /var/run/secrets/kubernetes.io/serviceaccount/ and fills the Configuration object.                                                                                                                                                          
                                                                                                                                                                                                                                              
  Without passing client_configuration explicitly, these functions modify a global                                                                                                                                                            
  singleton, which caused the bug in CI where both kubeconfig and in-cluster                                                                                                                                                                  
  credentials coexist.

You can find more details in the implementation of [load_kube_config](https://github.com/kubernetes-client/python/blob/582965c0a9612a76a52cf9e03d981db56a5c981c/kubernetes/base/config/kube_config.py#L754)

bsquizz (approved):

PR #582, opened 2026-06-10
feat: add install script and improve install docs

bsquizz:
## Summary
- Add a standalone `install.sh` script that creates a virtualenv and installs bonfire from PyPI, with colored output and Python version detection
- Simplify the non-interactive `bonfire version` output to print only the version string (e.g. `1.2.3` instead of `bonfire version 1.2.3`), making it easier to use in scripts
- Fix the version upgrade prompt to use `sys.executable -m pip` instead of bare `pip`, ensuring the correct Python interpreter is used
- Remove the outdated Homebrew install section from the README (formula is pinned to v5.11.0, latest release is v6.16.0)
- Reorganize README: move GitHub token setup into its own section

## Test plan
- [x] Run `bash install.sh` on a clean system and verify bonfire installs correctly
- [x] Run `bonfire version` in a non-interactive context (e.g. piped) and confirm it outputs only the version number
- [x] Run `bonfire version` interactively and confirm the rich output still works
- [x] Trigger the version upgrade check and confirm the suggested command uses the full Python path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ As an example, typing `bonfire deploy host-inventory` leads to the host-inventor
 
 - [About](#about)
 - [Installation](#installation)
+  - [GitHub token configuration](#github-token-configuration)
 - [MCP Server (AI Agent Integration)](#mcp-server-ai-agent-integration)
   - [Installing the MCP Server](#installing-the-mcp-server)
   - [Authentication](#authentication)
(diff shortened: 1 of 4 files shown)


mjholder (approved):
LGTM

PR #589, opened 2026-06-17
Fix nested spinner conflict causing LiveError

JuanmaBM:
### Problem
  Users reported crashes with `LiveError: Only one live display may be active at once` when running deploys. The error occurred because Rich does not allow multiple live displays (spinners) active simultaneously on the same console.

  ### Root Cause
  The code had nested spinners:
  - Line 1604: `status_spinner("Processing app templates...")`
    - Line 1015 (inside): `status_spinner(f"Fetching app deployment configs (source: {source})...")`

  Both spinners used the same console instance, which Rich rejects by design.

  ### Solution
  - **Removed** the nested inner spinner (line 1015)
  - **Enhanced** the outer spinner message to be more informative: `"Processing app templates from {source} ({app_list})..."`
  - Added additional `log.info()` calls to maintain visibility of the fetch operation progress

### DIFF
--- a/bonfire/bonfire.py
+++ b/bonfire/bonfire.py
@@ -1601,7 +1601,11 @@ def _cmd_config_deploy(
         clowd_env = None
 
     try:
-        app_list = ", ".join(app_names) if len(app_names) <= 3 else f"{', '.join(app_names[:3])}... ({len(app_names)} total)"
+        app_list = (
+            ", ".join(app_names)
+            if len(app_names) <= 3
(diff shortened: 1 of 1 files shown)


mshriver:
@JuanmaBM I dropped this change into my environment and deploy is running fine and the spinner is working as intended. Thanks!

JGiola (dismissed):

mshriver (dismissed):
Tested locally in my environment that produced the error, deploy worked fine with the spinner and messages visible as intended.

JGiola (approved):

